PR #5, opened 2019-10-21
Small typo fix for blower RB5015 in source (string) and documentation

limitz:
The documentation and source incorrectly use RM5015 as the name. The example still works because of the enumeration, but the code `blower(RM5015)` as shown in documentation doesn't.


nophead:
Thanks.

PR #10, opened 2020-01-10
Initial submission of E2020 aluminium extrusion.

martinbudden:
Here's my initial submission for E2020 aluminium extrusion. Once accepted I can add other types (E2040, E2060, E3030 etc) and add a submission for the `readme.md`.

I've tried to follow your coding style, but let me know if you'd like any changes.

I'm not sure what your line endings policy is, you don't have a `.gitattributes` file. My submission contains Windows style (`CRLF`) line endings, but I'm quite happy to change to Unix style (`CR`) - indeed that is my preference.

Extrusion dimensions are taken from [20x20 Aluminum Extrusion - 5 Series, Base 20 (MISUMI) (KHFS5-2020-1000](https://us.misumi-ec.com/vona2/detail/110302683830/?HissuCode=KHFS5-2020-1000)

### DIFF
--- a/lib.scad
+++ b/lib.scad
@@ -62,6 +62,7 @@ include <vitamins/spools.scad>
 include <vitamins/mains_sockets.scad>
 include <vitamins/ldrs.scad>
 include <vitamins/geared_steppers.scad>
+include <vitamins/extrusions.scad>
 
 use <vitamins/jack.scad>
 use <vitamins/meter.scad>
--- a/libtest.scad
+++ b/libtest.scad
@@ -32,6 +32,7 @@ use <tests/cable_strips.scad>
 use <tests/components.scad>
 use <tests/d_connectors.scad>
 use <tests/displays.scad>
+use <tests/extrusions.scad>
 use <tests/fans.scad>
 use <tests/fuseholder.scad>
 use <tests/geared_steppers.scad>
@@ -152,7 +153,7 @@ leadnuts_y = pillars_y + 40;
 pulleys_y = leadnuts_y +40;
 hot_ends_y = pulleys_y + 60;
 linear_bearings_y = hot_ends_y + 50;
-sheets_y = linear_bearings_y + 50;
+sheets_y = linear_bearings_y + 100;
 pcbs_y = sheets_y + 40;
 displays_y = pcbs_y + 150;
 fans_y = displays_y + 100;
@@ -201,6 +202,10 @@ translate([x0, pulleys_y])
 translate([x0, linear_bearings_y]) {
     linear_bearings();
     rods();
+}
+
+translate([x0+120, linear_bearings_y+30]) {
+    extrusions();
 }
 
 translate([x0 + 10, hot_ends_y])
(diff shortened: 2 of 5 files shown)


nophead:
Thanks. It has been on my todo list because I am planning to build a laser cutter out of extrusions.

Comments on the code: 

- There is already a fillet module here: https://github.com/nophead/NopSCADlib/blob/master/utils/fillet.scad#L25 , which will do 2D if h is zero. Having said that I would probably model the outside with rounded_square() and then subtract a mushroom shape rotated four times. The top of the mushroom has two radii which I haven't seen specified, so perhaps use the corner radius for those as well using hull. I am in Tenerife for winter otherwise I would attempt to measure it.

- Why did you wrap the vitamin in an assembly? I only do that when it comes with fasterners, etc.

nophead:
Regarding line endings, I am on Windows but my editor uses Unix line endings. This causes git to warn each time I commit a file. So I probably need a .gitattributes to sort that out.

martinbudden:
> Why did you wrap the vitamin in an assembly? I only do that when it comes with fasterners, etc.

Sorry, that was due to my ignorance. I'm still getting familiar with the code.

martinbudden:
> There is already a fillet module here: https://github.com/nophead/NopSCADlib/blob/master/utils/fillet.scad#L25 , which will do 2D if h is zero. Having said that I would probably model the outside with rounded_square() ...

I've updated to code to avoid using the fillet altogether. It's always good to avoid a `difference` since OpenSCAD is so slow...

nophead:
3D difference is slow but 2D difference is fast. In fact all the 2D ops
take no time at all, which is why I tend to model in 2D and extrude as much
as possible.

martinbudden:
> 3D difference is slow but 2D difference is fast. In fact all the 2D ops
> take no time at all, which is why I tend to model in 2D and extrude as much
> as possible.
> …

It's funny, when I first started using OpenSCAD, I used to draw things in 2D and then extrude them, then I changed to doing all the drawing in 3D, and fairly recently I've gone back to doing as much as possible in 2D, partly to make things faster, but also because the `thrown together` view in OpenSCAD also looks much better that way.

martinbudden:
OK, I'm fairly happy with the code now. Is there anything else you'd like changed?

If not, I'll do a rebase, so that this becomes just one commit to the master.

nophead:
Rather than rotate([0, 0, angle]) you can just use rotate(angle) for rotations around z.

martinbudden:
I've tidied the code and squashed the commits.

nophead:
Looks good. One comment though is if extrusion_corner() and extrusion_center_section() are only used in extrusion_cross_section(), and unlikely to be useful for anything else, then they can be nested inside extrusion_cross_section() to avoid polluting the global namespace.

Also vitamins/extrusions.scad should be included in lib.scad and a space for them found in libtest.scad so they appear in the main picture. To update libtest.png delete in and run tests.py. That should recreate it and also update the readme and make the image for it.

martinbudden:
I've added them next to the rods and linear bearings in the main picture.

martinbudden:
I've squashed and rebased again.

Since `libtest.png` and `readme.md` are generated files, they should be regenerated by you as part of accepting the pull request.

nophead:
OK, done. Thanks for your contribution.

martinbudden:
Great. Happy to contribute.

PR #15, opened 2020-01-11
Changed rail carriage coloring to be green with red endpiece

martinbudden:
This matches the common form and gives a good visual clue for the part.

### DIFF
--- a/tests/rails.scad
+++ b/tests/rails.scad
@@ -34,7 +34,7 @@ module rails()
             nut = screw_nut(screw);
             washer = screw_washer(screw);
 
-            rail_assembly(rail, length, rail_travel(rail, length) / 2);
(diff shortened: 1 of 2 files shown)


nophead:
This would need to be parametric because my rails are black, hence why I modelled them that way.

martinbudden:
I've parameterised the coloring of the carriage. I wasn't sure if you'd prefer it done by adding a parameter to the `carriage` `module`, or making the colors part of the property of the `carriage` type. I've done the former, but while doing it I thought the latter may be a better approach. What's your view?

nophead:
I think the way you have done it is best because otherwise it would require duplicating all the object definitions.

martinbudden:
OK. Anything else you'd like me to change in this?

nophead:
No it looks good, I will merge it after I have merged my own changes,

PR #17, opened 2020-01-12
Added ability to use microswitch on a pcb.

martinbudden:


martinbudden:
Withdrawn.

nophead:
May I ask why?

martinbudden:
Sorry the late reply. I felt that choosing the microswitch from the microswitches array was not really a very satisfactory solution - to susceptible to being broken by future changes.

nophead:
I see, yes I agree that is too brittle. I think you can just embed the object name in the PCB component list instead of an index.

PR #18, opened 2020-01-13
Changed brass color to standard definition.

martinbudden:
Changed brass color to standard as specified on several RGB generator sites, eg https://rgbcolorcode.com/color/brass and https://www.colorhexa.com/b5a642


nophead:
The problem is that my brass parts are not that dull. Currently I am using 
```brass                           = [255/255, 215/255, 0/255];```
because the new thread code needs to be able scale the colour, so they have to be numeric. The reason is the OpenSCAD lighting model make threads appear too bright because they tend to face directly to the lights.

**Comment by martinbudden** [comment]
Understood. However if a color is specified as brass in the library, then the user can legitimately expect it to be real brass. I suggest:

1. `brass` should be defined as the standard. The `[181/255,166/255,66/255]` form could be used so it is scalable.
2. You should define a new color, say `brassForThreads` and use that for your thread code.

There is also another aspect to this. The lighting model in OpenSCAD is not that good, however we can expect it to improve. This means that anywhere in the code that compensates for the lighting should be easily identifiable. Imagine that (in say 5 years time) the lighting was much better - then you could just change the value of your `brassForThreads` color. If you used the global brass for your threads and then altered this value in the future, then every design that used ordinary brass would be affected.

**Comment by nophead** [comment]
The HTML standard brass definition does not match any brass vitamins I own. It looks like brass that has aged, not brass inserts, brass nut, brass pillars, etc, which look a lot more golden. I have brass stock that colour but as soon as I machine it it looks bright and stays that way for years indoors.

I don't have a brassForThreads because threads can be any colour, I simply scale it down a little.

image

If OpenSCAD lighting improves I can just change the thread code to not scale it down.

**Comment by martinbudden** [comment]
Fair enough. My primary issue is that `brass` is distinguishable from the default yellow/gold that is used by OpenSCAD to render objects. In the current release of the NopSCADlib library `brass` is redefined as `"gold"` which makes brass inserts impossible to see.

**Comment by nophead** [comment]
Impossible to see against what? I colour all my printed parts in assembly views, so inserts are always clearly visible. When I am looking at a printed part on its own it may be shown in yellow but then there are no inserts in the view.

**Comment by martinbudden** [comment]
Here's and example
insertsexample

The inserts are only visible because in this case there is some clearance around them.

**Comment by nophead** [comment]
What is the material around them? Unless it is brass, gold or yellow plastic there would not be a problem. It should be coloured at the point the STL is included in the assembly.

Another way to avoid it is to use a difference colour scheme. The library uses Nature.

Incidentally the holes look too big for the inserts. The OD of the top of the insert is bigger than the recommended hole size which accepts the bottom part only until the rest is melted into the plastic.

**Comment by martinbudden** [comment]
The material is plastic, it has no inherent color - the color is chosen at the time of printing. I'm using the default color scheme, which now I have got used to I quite like. I don't want to change my color scheme, and I think there are probably plenty of other users who don't either.

I can get around the problem by overriding the color for the inserts, but the point I'm making is that the library shouldn't conflict with the OpenSCAD defaults.

(As for the size, the inserts fit fine in the printed part, I think in this case I have the inserts the wrong way around, I've only just converted to using your inserts, previously I was using my own private library which rendered inserts as cylinders, so I didn't necessarily get orientation right in all cases.)

**Comment by nophead** [comment]
I use ```pp1_colour``` to colour printed parts, which is lime. If I have adjoining parts I use ```pp2_colour``` and ```pp2_colour```, which are red and blue. Flexible parts use ```pp4_colour``` that is orange. When I print the parts I override the definitions locally with the actual colour I used if I want to get a more realistic render. E.g.
```
$pp1_colour = "grey";
$pp2_colour = "white";
```
at the top of the file.

martinbudden:
OK, revisiting this.

When I said I use the default colours, I didn't mean I let the colours default, what I meant was I actively choose `$pp1_colour = "gold";` (and `$pp2_colour = "orange";`), because, after experimenting with a few color schemes I decided the default one was the one I liked best. I'm afraid that I don't really like using "lime", "red" and "blue", or any of the more saturated colours.

I think that users setting `$pp1_colour = "gold";` is common enough for it to be catered for.

I understand your point about web "brass" is more an antique brass, and does not reflect brass parts.

How about setting `brass = [218/255, 190/255, 33/255];`? This is midway between antique brass and gold, so is distinguishable from "gold" parts and also better reflects real brass parts.

nophead:
This is what my brass inserts looks like:

image

This the current value equal to gold:

image

This is what you suggestion looks like:

image

I tried getting the pixel value from that image but because OpenSCAD doesn't do highlights it never looks bright enough. It looks more like flesh colour.

image

How about brass                           = [255/255, 220/255, 100/255];

image

nophead:
BTW, I do print lime and bright blue parts. I do have some gold PLA but never used it.

IMG_20171012_192229

IMG_20180511_094146

martinbudden:
[255/255, 220/255, 100/255] is ok for me.

Funnily enough saturated colors actually look quite good when printed, but for some reason I don't like them when in an OpenScad drawing.

nophead:
Fixed via https://github.com/nophead/NopSCADlib/commit/dcf258f11a14208eb13061388de7340f76040db0

PR #20, opened 2020-01-13
Add optional visible thread to studding

martinbudden:
OK, this one is a bit quick and dirty. However it was cheap to implement, costs nothing if not used and adds a reasonable visible thread to a lead screw. Try

`studding(8,100,2);`

to see if you like it.

I haven't looked at the spring code in any detail, but I imagine far fewer fragments could be used when using a spring as a thread. This would improve performance without really detracting from the visual effect.

### DIFF
--- a/vitamins/rod.scad
+++ b/vitamins/rod.scad
@@ -21,6 +21,7 @@
 //! Steel rods and studding with chamfered ends.
 //
 include <../core.scad>
+include <springs.scad>
 
 rod_colour = grey80;
 studding_colour = grey70;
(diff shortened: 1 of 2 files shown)


nophead:
Actually I am working on threads at the moment. They are dimensionally correct and printable if big enough. They are quite slow because the ends need to be intersected with a cylinder, so I am only enabling them when making documentation images.

image

image

nophead:
Also progress is slow because we are entertaining guests for the next fortnight,

martinbudden:
Looks great. A couple of comments:

1. It would be good to have a faster (and not necessarily dimensionally correct) form that is used when `$preview` is set.

2. Are you making them optionally user enabled? I wouldn't want them on for bolts, but I have a design which includes a lead screw and it would be good to have them switched on for that.

nophead:
Vitamins only get drawn at all when $preview is set. I have made printed threads and they do take longer to render under F6. They are done with sweep, so a would be a similar speed to a spring but to get the ends correct is slow. Similarly spring is slow if you choose the ground end option. 

What do the ends of your rods look like?

I have them enabled with a $variable that is set from an environment variable when running tests.py and views.py. It is set in tests/screws.scad and a few other files with threads but in user code, by default, it is disabled in the GUI.

martinbudden:
The ends of my rods were rubbish. The aim was not to get something accurate, but to give a good visual clue that the item was a lead screw rather than a rod. One of the downsides of OpenSCAD is that some of the renderings look a bit toy like and partly this is because small details are omitted.

I'd certainly like the option to set threads on for display for selected studding. I don't think for display  getting the ends correct is that important, since often the ends are obscured in a bearing or some other mounting.

PR #22, opened 2020-01-15
Added PCB buzzer.

martinbudden:
Added a small buzzer of type used on PCBs. Size and color are defaulted and parameterised, so they can be changed in the PCB array entry. (Size is settable, since I have seen at least two different sized buzzers used on PCBs.)

### DIFF
--- a/vitamins/pcb.scad
+++ b/vitamins/pcb.scad
@@ -200,6 +200,17 @@ module jack(cutout = false) { //! Draw 3.5mm jack
                 }
 }
 
+module buzzer(height, diameter, colour) { //! Draw PCB buzzer with specified height, diameter and color
+    color (colour)
+        linear_extrude(height)
+            difference() {
+                circle(d = diameter);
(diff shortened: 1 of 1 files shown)


nophead:
There isn't a test for this. Normally I only create PCB parts to put on a PCB so that PCB become the test case.

Perhaps we should have a dummy PCB with an example of each components.

martinbudden:
I'm using this on a RepRap Smart Controller PCB and also a BigTreeTech TFT 3.5 V3.0 controller PCB. I can submit a PR for either or both of those if you like.

A dummy PCB with an example of each component is a good idea.

nophead:
Yes please add those as they are useful to people want to make enclosures.

martinbudden:
I remember why I didn't submit those two - I need a potentiometer on the board for the RepRap Smart Controller, and a usb_A (single slot) for the TFT. I'll get those done and then submit the boards.

PR #23, opened 2020-01-15
Add ability to display filament on spool.

martinbudden:
Adds a `filament_depth` and `filament_color` parameters to spool. Adding filament can improve the look of a drawing.


nophead:
I made it more realistic but you can still get the solid version by setting filament_d to 0.

martinbudden:
Your version looks really good.

nophead:
The only problem is it tends to create a moire pattern. On an older version of OpenSCAD and an older computer I managed to enable anti-aliasing on the preview which was much nicer.

PR #24, opened 2020-01-16
Minor clarification of installation instructions.

martinbudden:


nophead:
What system is that on? I haven't set a PYTHONPATH on either Win7 or Ubuntu 18.

martinbudden:
I'm on Windows 10 and had to do this to get it to work.

I'll double check and report back.

martinbudden:
Sorry, my mistake. Just setting the `PATH` works fine.

However the instruction is a little bit ambiguous in this case: from the instructions I wasn't entirely clear whether you meant `PATH` or `PYTHONPATH` - perhaps you could make that explicit.

nophead:
I had never heard of PYTHONPATH before. It looks it is intended to extend the search path for imports, so you would set it if you had a library that you wanted to import. 

In this case we want it to find Python scripts to execute, rather than a library to impor, so I think setting PATH is correct on Windows and path on Linux. That is why I didn't mention it by name.

martinbudden:
Well, I was confused. I always find that it is the little bits of assumed knowledge that make using things difficult. I'm not a heavy python user, and I'm a recent returnee to Windows after a long time on Mac, so these little things trip me up.

nophead:
How about:
The ```NopSCADlib/scripts``` directory needs to be added to the executable search path, ```PATH``` on Windows and ```path``` on Linux.

martinbudden:
That's fine. (Unfortunately, I can't remember what it is on Mac. I think it's `path`, but I couldn't be 100% sure.)

nophead:
Fxied by https://github.com/nophead/NopSCADlib/commit/ba7e64233af4b0d3ede0b330e64f7f837c6131d2

PR #25, opened 2020-01-16
Added leadscrew. Used in stepper motors with long shaft.

martinbudden:
This time I've done a proper implementation which is reasonably performant. The thread is square-cut as on leadscrews. Here's an example of how it looks:

Screenshot (5)


nophead:
I am working a general thread library and and also using it for lead screws. Please don't do anymore threaded stuff until I have time to push it.

martinbudden:
OK, no worries.

PR #31, opened 2020-02-16
Added SCS bearing blocks, SCSnUU and SCSnLUU variants

martinbudden:
Dimensions taken from  https://www.makeralot.com/scs20uu-linear-ball-bearing-block-p11/

### DIFF
--- a/vitamins/scs_bearing_block.scad
+++ b/vitamins/scs_bearing_block.scad
@@ -133,7 +133,7 @@ module scs_bearing_block_hole_positions(type) {
                 children();
 }
 
-module scs_bearing_block_assembly(type, sheet_thickness, screw_type, nut_type) { //! Assembly with screws and nuts in place
+module scs_bearing_block_assembly(type, part_thickness, screw_type, nut_type) { //! Assembly with screws and nuts in place
 
     scs_bearing_block(type);
 
@@ -143,7 +143,7 @@ module scs_bearing_block_assembly(type, sheet_thickness, screw_type, nut_type) {
     washer_thickness = washer_type ? washer_thickness(washer_type) : 0;
 
     G = scs_block_side_height(type);
-    nut_offset = G + sheet_thickness + nut_thickness(nut_type) + washer_thickness;
+    nut_offset = G + part_thickness + nut_thickness(nut_type) + washer_thickness;
     screw_length = screw_longer_than(nut_offset);
 
     scs_bearing_block_hole_positions(type) {
--- a/vitamins/scs_bearing_block.scad
+++ b/vitamins/scs_bearing_block.scad
@@ -23,8 +23,10 @@
 
 
 include <NopSCADlib/core.scad>
-include <NopSCADlib/vitamins/screws.scad>
-include <NopSCADlib/vitamins/linear_bearings.scad>
+use <NopSCADlib/vitamins/screw.scad>
+use <NopSCADlib/vitamins/nut.scad>
+use <NopSCADlib/vitamins/washer.scad>
+use <NopSCADlib/vitamins/linear_bearing.scad>
 
 function scs_size(type)                 = [type[4],type[6],type[5]]; //! Size of scs bracket bounding block
 function scs_hole_offset(type)          = type[2]; //! Offset of bearing hole from base of block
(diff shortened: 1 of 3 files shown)


vitamins/scs_bearing_block.scad
```diff
@@ -0,0 +1,147 @@
...
+// If not, see <https://www.gnu.org/licenses/>.
+//
+
+//
+// SCSnUU and SCSnLUU bearing blocks
+//
+
+
+include <NopSCADlib/core.scad>
+include <NopSCADlib/vitamins/screws.scad>
```
nophead: Vitamins only need to be included if an object name is needed, which is not the case here because screws and bearings come from the scs_bearing object.   It is better to use screw.scad, nut.scad and linear_bearing.scad.

vitamins/scs_bearing_block.scad
```diff
@@ -0,0 +1,147 @@
...
+function scs_bearing(type)              = type[14]; //! Linear bearing used
+
+
+sks_bearing_block_color = grey90;
+
+module scs_bearing_block(type) { //! Draw the specified SCS bearing block
+    vitamin(str("scs_bearing_block(", type[0], "): ", type[0], " bearing block"));
+
+    T = type[1];
+    h = type[2];
```
nophead: These should use the accessor functions so the offset is defined in two places that have to match.

vitamins/scs_bearing_block.scad
```diff
@@ -0,0 +1,147 @@
...
+
+    screw_separation_x = scs_screw_separation_x(type);
+    screw_separation_z = scs_screw_separation_z(type);
+
+    for(x = [-screw_separation_x, screw_separation_x], z = [-screw_separation_z, screw_separation_z])
+        translate([x / 2, G - h, z / 2])
+            rotate([-90, 0, 0]) {
+                screw(screw_type, screw_length);
+                translate_z(-screw_length  + washer_thickness)
+                    nut(screw_nut(screw_type))
```
nophead: There is a nut_and_washer(nut) module.

vitamins/scs_bearing_block.scad
```diff
@@ -0,0 +1,147 @@
...
+                                translate([0, y])
+                                    circle(r = boltHoleRadius);
+                        }
+    }
+    not_on_bom() no_explode()
+        linear_bearing(scs_bearing(type));
+}
+
+
+module scs_bearing_block_assembly(type, screw_length, screw_type) { //! Assembly with screws and nuts in place
```
nophead: I don't understand why the screw length is specified or why the nuts are placed a washer's thickness from the end of the screw. All other assemblies like this are passed a panel thickness. The screw length is then calculated and the nut and washer would be offset downwards by the panel thickness.

vitamins/scs_bearing_block.scad
martinbudden: OK.

vitamins/scs_bearing_block.scad
martinbudden: Yes, I've recently discovered that.

nophead:
I noticed the same comments apply to sk_bracket_assembly().

Another point is that I always factor out mounting hole positions into a module that positions its children. That can be used to drill mounting holes in a panel as well as place the fasteners in an assembly, or make holes in the part itself without repeating code.

vitamins/scs_bearing_block.scad
martinbudden: Screw length is specified so that the bearing block can be attached to items of varying thickness. I hadn't thought of passing the panel thickness instead - I'll do that.

martinbudden:
Yes, that's a good way of doing things. I'll factor the hole positions into a separate module.

vitamins/scs_bearing_block.scad
nophead: A simple example here: https://github.com/nophead/NopSCADlib/blob/master/printed/door_hinge.scad#L159

nophead:
BTW I removed the let($show_threads=true). That should only be set in tests where the object has a thread itself, so it can be seen that it works. The bearing blocks only use screws that will have threads if the SHOW_THREADS environment variable is set, which it is when I run the tests.

PR #32, opened 2020-02-16
Added space before screw length when length < 10 to ensure correct ordering

martinbudden:
So in the BOM we get:
```
Screw M3 dome x  6mm
Screw M3 dome x  8mm
Screw M3 dome x 10mm
```
rather than
```
Screw M3 dome x 10mm
Screw M3 dome x 6mm
Screw M3 dome x 8mm
```


nophead:
Clever trick! Sorts out the order but doesn't show up in the HTML.

PR #38, opened 2020-02-23
Added scad/printed to stl search path.

martinbudden:
Allows better organisation for a largish build. Main assembly and sub assemblies can be in `scad`, printed parts in `scad/printed` and vitamins in `scad/vitamins`.


nophead:
Falls over if the printed director does not exist.

martinbudden:
Ooops.

Any chance you could make the fix? My python is a bit rusty and while I could make the fix, it would probably take me a while to figure it out, whereas I imagine you could do it off the top of your head.

nophead:
OK I will fix it

martinbudden:
Thanks.

PR #39, opened 2020-02-23
Internal and external corner brackets for 20nn extrusion.

martinbudden:
Updated, rebased and ready to merge.

### DIFF
--- a/tests/extrusionBracket.scad
+++ b/tests/extrusionBracket.scad
@@ -1,0 +1,61 @@
+//
+// NopSCADlib Copyright Chris Palmer 2018
+// nop.head@gmail.com
+// hydraraptor.blogspot.com
+//
+// This file is part of NopSCADlib.
+//
+// NopSCADlib is free software: you can redistribute it and/or modify it under the terms of the
+// GNU General Public License as published by the Free Software Foundation, either version 3 of
+// the License, or (at your option) any later version.
+//
+// NopSCADlib is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+// without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along with NopSCADlib.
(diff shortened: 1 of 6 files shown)


nophead:
This is missing a description comment.
Camel case isn't the convention in this library, so the filenames should be extrusion_bracket.scad.
Why isn't this parametric, these brackets are available for other extrusion sizes?

martinbudden:
I'll add the comment.

Sorry about the camelCase, just force of habbit.

They are not parametric because the dimensions are not available and I don't have any e30 or e40 brackets in my possession. These are certainly not items that follow a standard - I believe the external corner brackets are actually furniture brackets that have been co-opted. Internal corner brackets are available for e30 extrusion, but it is possible they are an entirely different design.

nophead:
When I ordered my 2020 internal brackets they were available for 30 and 40 as well and the picture didn't change when I selected different sizes, so I assume they are just bigger. 

It isn't necessary to add the other ones if you don't have the sizes but I think the dimensions should come from an object list so that the bigger ones can be added easily, as there are clearly more than one size.

It doesn't matter too much how much comes from the object and how much is hard coded to start with.  It might turn out they differ in more ways than anticipated but then more fields can be added to the object and it is completely backwards compatible. This is generally how things evolve, like the T-nuts.

I see some are described as "furniture brackets" but they have keys that fit into the extrusion slot so I think they are only use with extrusions. Possibly it is a mistranslation from Chinese.

martinbudden:
OK, I can parameterise them. I'd like to minimise the number of parameter used - I don't really like to generalise based on just one example. It's difficult to know what is variable, and what is derived from other variables. Ie I may have constants `a`, `b`, `c` and `d`, but not realise that (say) `d` is actually a derived value, ie something like `c-a`. When we add new variants it will be easy to pull constants out of the function into the properties array.

nophead:
Yes that is fine. If the design starts out as recognising it is parametric and passes an object list then it doesn't matter what is in the list, what is hard coded, or comes from a function as long as the function is passed the list. Information can be moved around as new sizes are added without breaking the interface.

PR #40, opened 2020-02-23
Tidy of sk_brackets as per scs_bearing_blocks code review.

martinbudden:


nophead:
Actually  I have started merging this and ended up changing the t-nut stuff. I made the Z origin where the nut meets the extrusion and removed the stem from the thickness to simplify nut placement and screw length calculation. I also removed washers from t-nuts.

martinbudden:
Yes, my thoughts as well. I was planning making those changes as well, but it looks like you got there before me.

nophead:
OK I included your new test and merged my version. So I will close this.

I am thinking it might need to be screw_shorter_than when using t-nuts as there isn't a lot of room below the nut in an extrusion.

martinbudden:
Yes, I've already changed to using `screw_shorter_than` in some of the PRs. Once everything is merged in I'll double check everything.

martinbudden:
OK, I don't think you've done the tNut changes in the optimum manner. I think the thickness should be the true thickness of the nut, I'm not generally in favour of 'lying' about attributes.

Also the tNut thickness is used by `extrusionSlidingNut` which is used in `extrusion20InnerCornerBracket` which is now broken.

My plan was to flip the nut and offset it by the stem height, which should make placement easy with having to lie about dimensions.

PR #41, opened 2020-02-24
Improved tnut positioning and hammer nut handling.

martinbudden:
I think this is the way to handle the tnut positioning. tnut is drawn in correct orientation and offset for screw, no faking its thickness or using nyloc functionality.

Also hammer nuts are drawn using `sliding_t_nut` so that substituting hammer nuts for tnuts is facilitated.

### DIFF
--- a/tests/nuts.scad
+++ b/tests/nuts.scad
@@ -58,10 +58,10 @@ module nuts() {
 
         translate([0, 80]) {
             if(n == M3_nut)
-                hammer_nut(M3_hammer_nut);
+                sliding_t_nut(M3_hammer_nut);
 
             if(n == M4_nut)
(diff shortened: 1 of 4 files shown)


nophead:
I don't think this works because the stem of the nut is below the mating surface, so should not be included in the thickness when calculating the screw length. That is generally what nut_thickness() is used for.

Using the nyloc field is a bit of a hack but with the new, un-named field you can't get the total height if that was ever needed. Since the stem dangles in the slot of the extrusion without meeting anything I don't think the total height is ever needed, other than to draw it.

martinbudden:
OK.

I'll make a new submission that unifies the sliding tnuts and the hammer nuts.

Withdrawn.

PR #47, opened 2020-02-26
Added variants for the KP08 pillow blocks

Drachenkaetzchen:
This patch adds both 15mm and 18mm variants of the pillow blocks.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1131,7 +1131,8 @@ KP pillow block bearings
 | ---:|:--- |:---|
 |   1 | ```kp_pillow_block(KP000)``` |  KP000 pillow block |
 |   1 | ```kp_pillow_block(KP001)``` |  KP001 pillow block |
-|   1 | ```kp_pillow_block(KP08)``` |  KP08 pillow block |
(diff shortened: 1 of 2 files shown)


nophead:
This didn't build, I had to fix the test routine. Please run the test before submitting PRs.

I have merged it but I think it highlights a bug in the drawing. There are gaps between the top of the webbing and the ring around the bearing. I think the real objects won't have gaps.

martinbudden:
I was a little surprised to see that only the H parameter was changed in this PR. Generally the parameters are interrelated, so changing one will require changes to others.

I'll have a look at this and submit a fix in due course.

Drachenkaetzchen:
> I was a little surprised to see that only the H parameter was changed in this PR. Generally the parameters are interrelated, so changing one will require changes to others.

I was surprised too when I had a look at several "datasheets" - everything else except the hole offset seems exactly the same, including the overall height of the pillow block.

> I'll have a look at this and submit a fix in due course.

Thanks!

martinbudden:
OK, I've checked this. There is nothing wrong with Drachenkaetzchen's changes to the H parameter, however in my code the webbing is not drawn entirely correctly. I'll fix that.

PR #51, opened 2020-02-27
Defaulted extrusion center to true.

martinbudden:


nophead:
I think you also need to update all the places it is used in the tests.

martinbudden:
Should be OK now.

nophead:
Is cornerHole something you buy the extrusion with or something you drill yourself? I.e. should it go on the bom?

I have been integrating your new parts into the big picture generated by libtest.scad, so I have changed the center default myself and changed most of the calls. I haven't changed the extrusion test itself as similar parts are shown centred. It doesn't actually affect the picture because the axes are not shown. It does affect the position in the big picture though.

Also center doesn't need to go in the vitamin example call. I only print the args that affect the bom, i.e. what the vitamin is, not how it is positioned. That is why I ask about cornerHole.

At some point in the future I may generate thumbnails for the vitamins from the example call.

nophead:
Implemented by https://github.com/nophead/NopSCADlib/commit/6470f1b0045bae8bc9699b7e1523f1f664e2dd8b, so closing this one.

PR #52, opened 2020-02-27
Added BigTreeTech TFT35v3.0

martinbudden:
This does not contain all the components on the PCB, but all the sockets and everything around the edges are included.

Dimensions taken from https://github.com/bigtreetech/BIGTREETECH-TFT35-V3.0/blob/master/TFT35%20V3.0%20user%20manual-2019-11-28.pdf where available and otherwise measured by me and cross checked against https://www.thingiverse.com/thing:3818344

I've used JST_XH sockets they have the same pin spacing but slightly different plastic surrounds from the sockets on the TFT35.

And finally, in the test, it looks a bit odd with the standoffs.  Should the test be changed not to use standoffs (they don't really apply to the RAMPS endstop either)?

### DIFF
--- a/vitamins/pcbs.scad
+++ b/vitamins/pcbs.scad
@@ -351,7 +351,36 @@ RAMPSEndstop = ["RAMPSEndstop", "RAMPS Endstop Switch",
     ],
     []];
 
+BigTreeTech_TFT35v3_0 = ["BigTreeTech_TFT35v3_0", "BigTreeTech TFT35 v3.0",
+    110, 55.77, 1.6, 0, 3, 0, "green",  false,
+    [ [-3.12, 3.17], [-3.12, -3.17], [3.12, -3.17], [3.12, 3.17] ],
+    [
+        // mock up an LCD screen
+        [  61,  28,    0, "-chip", 84, 54.5, 4, grey70 ],
+        [  61,  28,    0, "-chip", 80, 50, 4.1, "#0000cc" ],
+        [  10, 7.5,    0, "-button_6mm" ],
+        [  10,45.5,    0, "-buzzer", 4.5, 8 ],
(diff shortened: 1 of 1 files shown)


nophead:
I think it should be a display rather than a PCB. They have an attached PCB but they get shown in the display list, without standoffs.

Nor sure about the endstop, how is it normally mounted? 

I normally put the smallest PCBs that have components first in the list to get a better view. At some point I will probably put the perfboards on a second row as it is getting a bit long.

The test PCB would look better if it was more square as the picture would get scaled less.

martinbudden:
I'll have a look at making it into a display.

I think normally people print some kind of enclosure for it - there are certainly lots of them on thingiverse.

Yes, the PCB row is getting a bit long. :-)

nophead:
Yes display object is designed to be able to make an enclosure around it. 

image

Regarding the endstop PCB. If it is mounted on plastic parts then pcb_assembly should be able to make sensible standoffs. There is currently a bodge for Arduino that makes tapered pillars because there is no proper clearance. It is triggered by using undersized screws, which you have to do with Arduino because there is no room for the screw heads. The holes are big enough for M3 but you can only just fit M2.5 screw heads.

On the endstop the screws obstruct the connector so I don't know what is going on there.

Interesting it uses lever endstops, they reduce the repeatability, I always use buttons.

nophead:
Looking at pictures on line it looks like the two screws holes near the connector are not for mounting. I think they are for a cable tie over the wires or something so I added a [bodge](https://github.com/nophead/NopSCADlib/commit/b56ddea1e3358ef03cd342db1cff0e913ae5742d) to not put screws in them.

martinbudden:
Superseded by https://github.com/nophead/NopSCADlib/pull/97.

PR #66, opened 2020-03-15
Fix handling of (Windows) paths with spaces

FLamparski:
I was getting errors when processing paths like `C:\Users\ftwie\Documents\Fisher Price CNC\scad` - the script just wouldn't pick up directories where the path had a space character because it didn't properly unescape the `\ ` sequence. This patch fixes that. Only tested with Windows paths.


nophead:
Thanks. I always avoid paths with spaces as it usually causes problems like these.

PR #67, opened 2020-03-15
Adds DIN 562 square nuts

FLamparski:
These are the "thin" square nuts used eg. in the Prusa i3 MK3


nophead:
Thanks for the addition. I fixed the threads, they need to be added, not subtracted.

PR #72, opened 2020-04-09
nut_radius is needed for screw_boss_diameter

ledvinap:
using `include <NopSCADlib/lib.scad>` in main file fails in `screw_boss_diameter`:
`WARNING: Ignoring unknown function 'nut_radius', in file ../../../sw/OpenSCAD/libraries/NopSCADlib/vitamins/screw.scad, line 41.`

Maybe my library usage is wrong, I did not investigate further ...

### DIFF
--- a/vitamins/screw.scad
+++ b/vitamins/screw.scad
@@ -23,6 +23,7 @@
 include <../utils/core/core.scad>
 
 use <washer.scad>
+use <nut.scad>
 use <../utils/rounded_cylinder.scad>
 use <../utils/thread.scad>
 include <inserts.scad>


nophead:
Are you actually calling screw_boss_diameter()?

It isn't used anywhere in NopSCADlib, so there is no test coverage, which is why nobody noticed it. 

The function originally came from Mendel90 and I haven't used it from NopSCADlib myself.

ledvinap:
I needed dimension to sink screw head with washer ... `screw_boss_diameter` seemed appropriate, mainly because my poor technical english. Using washer size is of probably correct way to do it.

nophead:
OK I was just curious. I haven't used that function because I now use brass inserts instead of nut traps in most of my designs.

PR #75, opened 2020-06-16
Added vacuum flanges

KOLANICH:
from [PR to BOLTS](https://github.com/boltsparts/BOLTS/pull/169)

### DIFF
--- a/vitamins/vacuum/CF-Flange.scad
+++ b/vitamins/vacuum/CF-Flange.scad
@@ -1,0 +1,105 @@
+/*
+CF vacuum flanges for OpenSCAD
+This is free and unencumbered software released into the public domain.
+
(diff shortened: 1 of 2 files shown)


nophead:
This is completely incompatible with NopSCADLib. I don't represent objects the same way. Objects are represented as lists of properties, not default parameters to a module.

Also I don't make hollow tubes by differencing cylinders as it is very slow.

nophead:
Take a look at how a similar shaped part is modelled in NopSCADlib https://github.com/nophead/NopSCADlib/blob/master/vitamins/leadnut.scad.

If you submit something coded in the same coding style, with its own test file and the correct comment markup to make the documentation in the readme I will consider merging it.

PR #82, opened 2020-11-08
Added optional jst connector to stepper motors.

martinbudden:
Only added it to the NEMA17 motors, since I don't have any of the other types to confirm correctness.

### DIFF
--- a/tests/stepper_motors.scad
+++ b/tests/stepper_motors.scad
@@ -24,7 +24,7 @@ use <../utils/layout.scad>
 module stepper_motors()
     layout([for(s = stepper_motors) NEMA_width(s)], 5) {
         rotate(180)
-            NEMA(stepper_motors[$i]);
+            NEMA(stepper_motors[$i], 0, $i > 1 && $i < 5);
 
         translate_z(4)
             NEMA_screws(stepper_motors[$i], M3_pan_screw, n = $i, earth = $i > 4 ? undef : $i - 1);
--- a/vitamins/pin_headers.scad
+++ b/vitamins/pin_headers.scad
@@ -25,7 +25,7 @@
 //                                                              c
 //
 2p54header    = ["2p54header",   2.54, 11.6, 3.2, 0.66, "gold", grey(20), 8.5];
-jst_xh_header = ["JST XH header",2.5,  10,   3.4, 0.64, "gold", grey(90), 7];
+jst_xh_header = ["JST XH header",2.5,  10,   3.2, 0.64, "gold", grey(90), 7];
 
 pin_headers = [ 2p54header ];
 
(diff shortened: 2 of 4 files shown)


nophead:
Thanks, something on my todo list.

The only motors on the list that have connectors are the NEMA17M and NEMA17M8 sizes that I got from RoboDigg. The larger and smaller ones I posses just have wires. What sizes are you using?

I would be happy to add connectors to just those two but if you want the larger and smaller motors to have connectors we will need to have two variants of them.

martinbudden:
I've just got NEMA17 motors. Some with wires and some with JST connectors. I don't think we need two variants of those - if the motor with the JST connector fits, then so will the one with wires.

As for the other sizes, I think a good approach is only to add a variant with a connector when you get one in your possession, so that you can check the dimensions.

nophead:
Perhaps it should be a parameter of the module rather than a property of the object to avoid having two variants then.

martinbudden:
Yes, that's a better idea. I'll update the PR.

nophead:
The connector seems to be offset from the tab.

image

nophead:
I fixed it by using the straight version of the connector and the width of the connector / 2.

martinbudden:
The gap between the connector and tab is deliberate - on stepper motors there is a tab of a circuit board there, which I didn't bother drawing. By moving the connector closer to the tab you have now made it dimensionaly incorrect.

nophead:
OK I will add the PCB then. Its hard when I don't have any parts with me.

Do you know how they stop the pins shorting to the metal. Is there a gap and some insulation or is the tab hollow?

nophead:
Looks like there is no standard position for this. Here are the dimensions of the 40mm motors I have.

image

It sticks out further and is mounted higher than your version. Do you have an alternate drawing?

martinbudden:
I made this using dimensions I measured rather than from a drawing. I guess it is good enough at the moment. Having the connector ensures there are no nasty surprises when assembling and realising there is no room for the connector.

nophead:
I found out the connector is actually smaller because it is a JST PH series with 2mm pitch. I will attempt to model that and make it match the drawing as that is the only NEMA17 motor I have with a connector. The metal tab is a lot thicker, so I assume you must have a different motor.

martinbudden:
It seems like the size of the tab is not standardised. I think the important thing is to slightly oversize rather than undersize, that way if the motor fits in the drawing it is more likely to fit in actuality.

nophead:
I think the tab thickness only depends on the end cap height, so I have made that parametric and set it correctly on the NEMA17 I have that has a connector. Basically the connector butts up the laminations and the tab is below that by the connector height and the standard thickness of a PCB, 1.6mm, even though it looks like some PCBs are thinner.

I made the JST connector more parametric and modelled the PH connectors used by the motors. In doing so I made the pin offsets of the XH connectors more accurate. If you have any on a PCB you with need to move vertical ones by 0.525mm and right angle ones by 0.4mm as connectors have their origin on the pin line.

image

Issue #81, opened 2020-11-08
The `tests.py` script is hardcoded to NopSCADlib, making it not generally useful.


martinbudden:
I've got a project with some tests in the `tests` directory. `tests.py` cannot currently be used to run them, but a few small changes could rectify this. In particular:

1. `tests.py` does not search the project `tests` directory. Adding `tests/` to the `locations` array fixes this.
2. The output is hardcoded to `readme.md`, which means it overwrites the output from `make_all.py`.
3. `libtest.scad` is required for `tests.py`. This is not required in the general case.
4. The heading blurb is directly encoded in `tests.py`. It would be better if it was scraped from `libtest.scad` - this would have the added benefit of making it easier to maintain the blurb.

nophead:
Well yes it is designed specially to test the library and make its documentation. I don't have any tests in my projects other than building them makes their documentation and I see when stls and assembly views change unexpectedly, which is really only what tests.py does for the library.

I don't have anything resembling the library structure in my projects. For example it assumes vitamins, utils and printed directories and it produces a contents section split on three hard coded letters of the alphabet.

I don't have properly coded vitamins in my projects. Most projects only have one file in them and get everything from the library. I do mock up a few vitamins specific to a project but don't bother to parameterise or document them unless I want to use them in more than one project and then I move them to the library and make them more general.

What do you have in you projects that needs testing that isn't covered by make_all.py?

martinbudden:
There are a few reasons I have tests that suplement `make_all.py`:

1. `make_all.py` takes ages to run if the project is largish.
2. I do have vitamins in my project. From your perspective, if you create a new vitamin, you might as well put it in the library. From my perspective, the vitamin might not be generally useful, or you might not want it in the library for some reason.Or the vitamin might be "rough and ready", that is its dimensions are correct, but it lacks colouring and decoration and I might not have the time to get it to library standard.  So I have test files to test vitamins. 
3. I can create test files for "unnatural" subassemblies. That is subassemblies that don't actually occur in the project, but that are created to test fit and alignment without other objects getting in the way. Also "naked" subassemblies can be useful, that is subassemblies that don't have the outside walls, or the structural elements (such as extrusions).

I managed to get `tests.py` working for my purposes. There are very few changes required and I think they might be useful to others.

nophead:
Yes make_all gets very slow as projects get larger. I am hoping the persistent cache addition to OpenSCAD will help there.

[...]

PR #83, opened 2020-11-08
Updated tests.py to better support generic testing.

martinbudden:
Partial fix to https://github.com/nophead/NopSCADlib/issues/81 .

Could you add a change to allow the filename to be specified on the command line - that would be sufficient to make this usable for me.

### DIFF
--- a/scripts/tests.py
+++ b/scripts/tests.py
@@ -90,6 +90,8 @@ def usage():
     sys.exit(1)
 
 def tests(tests):
+    doc_base_name = "readme"
+    doc_name = doc_base_name + ".md"
     scad_dir = "tests"
     deps_dir = scad_dir + "/deps"
     png_dir  = scad_dir + "/png"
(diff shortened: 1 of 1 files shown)


nophead:
I don't understands this PR. It breaks the library because tests is where the test files live. It isn't a category of things to be tested.

I get this:

image

What do you have in the test directory? Tests or things to be tested?

martinbudden:
Ah, I see. I sort of assumed that everything in tests should be tested. I'll come up with a better PR.

nophead:
I am working on it at the moment. I have used the presence of libtest.scad to decide if it is a library build or not.

Tests contains just the test code. It then looks in vitamins, utilities and printed to find the source code and uses that to categorise them in the table of constants. I think you need to put your code that is to be tested in the folders and the test code in test to mimic the library structure to make it work.

martinbudden:
OK, that sounds good. I'll leave you to it.

PR #90, opened 2020-11-13
Add optional bowden connector to E3D hotends.

martinbudden:


nophead:
Thanks for this.
I got an error because you passed "silver" to the thread and it needs the numerical version silver. I guess you must be using an older version of OpenSCAD if you didn't get an error. Recent snapshots give warnings for mixing strings and numbers.

The Bowden connectors are actually pneumatic push-in connectors available in different sizes and colours so they will probably become another parametric object at some point. I have lots of pneumatic parts to model eventually. Perhaps when I build a pick and place machine.

Issue #91, opened 2020-11-13
Belt gap does not allow a belt to be offset with itself at the gap positon.


martinbudden:
This is required, for example in certain forms of coreXY. See image, the current belt gaps are highlighted in pink.
![CoreXY](https://user-images.githubusercontent.com/194586/99070519-4c505a80-25a8-11eb-928a-043c75cedb8e.png)

Fix will be submitted soon.

nophead:
Interesting. Is that core XY with two separate planar belts? I thought it had one belt in a spiral normally. I wouldn't like to model that so use H-BOT myself! I don't think the racking force matters with trapezoidal rails. I have a milling machine with a single Y axis ball screw in the middle. I have yet to test my H-BOT though.

martinbudden:
Yes, that is coreXY. CoreXY always uses two belts, although there are different ways they can be offset from each other. It's not that difficult to model. Here's the top view with the fix (and you can see why I wanted the JST connectors on the motors).
![CoreXYfix](https://user-images.githubusercontent.com/194586/99071865-db5e7200-25aa-11eb-97b1-23324e2150e6.png)

martinbudden:
And for your interest, here's current state of the project:
![HIGLR_main](https://user-images.githubusercontent.com/194586/99072878-c4208400-25ac-11eb-938e-a0bf52bed06a.png)

nophead:
Not sure what is happening to your latest commit to the PR. I see this:

![image](https://user-images.githubusercontent.com/566149/99115673-b853b280-25ea-11eb-9d77-62e52cf5d290.png)

If I click on the commit number in the email I get this:

![image](https://user-images.githubusercontent.com/566149/99115815-ecc76e80-25ea-11eb-8abf-08629e7619be.png)

martinbudden:
I think I made a mistake when rebasing. It should be ok now.

nophead:
Seems there are two ways of doing coreXY. From the Reprap wiki:

>There is two ways to built the CoreXY belt system.
>
>Both belts are on same plane so belts are crossing and they tend to rub one against the other at crossing point. Twisting the belt on the crossing path limits the rubbing. A small finger can be installed to separate the belts at crossing point.
>The two belts could be set on two different levels (see image). This creates a small tilting momentum on the hotend carriage and increase the momentum on shafts if pulleys are installed on cantilevered shafts. This also save some space on the return pulley side as the pulleys could be stacked.

[...]

PR #92, opened 2020-11-13
Allow user to set y size of belt gap.

martinbudden:
Fixes https://github.com/nophead/NopSCADlib/issues/91

### DIFF
--- a/vitamins/belt.scad
+++ b/vitamins/belt.scad
@@ -54,7 +54,7 @@ module belt(type, points, gap = 0, gap_pos = undef, belt_colour = grey(20), toot
 
     tangents = rounded_polygon_tangents(points);
 
-    length = ceil((rounded_polygon_length(points, tangents) - gap) / pitch) * pitch;
+    length = ceil((rounded_polygon_length(points, tangents) - (is_list(gap) ? gap.x + gap.y : gap)) / pitch) * pitch;
 
     module shape() rounded_polygon(points, tangents);
 
@@ -65,7 +65,7 @@ module belt(type, points, gap = 0, gap_pos = undef, belt_colour = grey(20), toot
             translate([gap_pos.x, gap_pos.y])
                 rotate(is_undef(gap_pos.z) ? 0 : gap_pos.z)
                     translate([0, ph - thickness / 2])
-                        square([gap, thickness + eps], center = true);
+                        square(is_list(gap) ? [gap.x, gap.y + thickness + eps] : [gap, thickness + eps], center = true);
 
     color(belt_colour)
         linear_extrude(width, center = true)


vitamins/belt.scad
```diff
@@ -64,8 +64,8 @@ module belt(type, points, gap = 0, gap_pos = undef, belt_colour = grey(20), toot
         if(gap)
             translate([gap_pos.x, gap_pos.y])
                 rotate(is_undef(gap_pos.z) ? 0 : gap_pos.z)
-                    translate([0, ph - thickness / 2])
-                        square([gap, thickness + eps], center = true);
+                    translate([0, is_list(gap) ? -th / 4 : ph - thickness / 2])
```
nophead, line 67: Why tooth height / 4?

martinbudden: That centers it on the gap.

nophead: I don't see how. The belt is modelled relative to the pitch line, so adding the pitch height and subtracting half the thickness gets to the middle of the belt, so a rectangle slight thicker thicker than the belt just covers it.

How is the tooth height relevant and why divide it by 4?

martinbudden: I'll look into it again.

nophead: If gap.y is the offset from one end to the other that I think you need to add the belt thickness + eps to that to cover both. And the centre would be ph - thickness / 2 + gap.y / 2, I think.

martinbudden: OK, I think it should be:
```
                    translate([0, (ph - thickness) / 2]);
```
ie, I think your original value of 
```
                    translate([0, ph - thickness / 2]);
```
is incorrect.

But in any case, the value does not need to change when a y value for the gap is set. ie the full code for `gap` is
```
    module gap()
        if(gap)
            translate([gap_pos.x, gap_pos.y])
                rotate(is_undef(gap_pos.z) ? 0 : gap_pos.z)
                    translate([0, (ph - thickness) / 2])
                        square(is_list(gap) ? [gap.x, gap.y]: [gap, thickness + eps], center = true);
```
But in any case, the difference is marginal, so I won't argue against it staying as `ph - thinkness / 2`.

nophead: Why would you divide ph by 2? The pitch height is a magic number part way through the teeth measured from the tips of the teeth. Adding that gets to the tips and subtracting half the total thickness gets back to the centre, for a square centred on the belt. It looks spot on to me.

image

martinbudden: OK, that makes sense. I had a misunderstanding of the pitch height.

nophead: So how is the user specified y gap centred? Or do you pass in a length twice the actual gap?

nophead: I think the correct y gap needs to be passed in because it also needs to be subtracted from the length.

martinbudden: The y cap is a static property of the belts layout. It is the belt separation in the y axis plus the belt thickness. It certainly needs to be passed in as a parameter.

nophead: So if you pass the actual y gap how is the square to remove it working without adding half the gap to centre it correctly?

And gap.y also needs to be subtracted from the length added to the BOM.

martinbudden: Not sure what you mean about centering the gap, the call is:
```
square(is_list(gap) ? [gap.x, gap.y] : [gap, thickness + eps], center = true);
```
that is the center parameter is set to true.

As for the length of the belts, yes it needs to be subtracted, I forgot that.

nophead: Yes I guess you just pass the centre of the gap in the ```gap_pos```. 

Does the y value you pass in need to be inflated by the belt width as it is from pitch line to pitch line? 

So I think you subtract gap.x + gap.y from the length but need a square gap.y + thickness + eps to remove it.

martinbudden: Yes, the y value needs to be inflated by the thickness. In my code I pass in the the `ySeparation+2` just to be sure, but that would result in an incorrect belt length. So I think the parameter passed in should be just the separation (not inflated by the thickness) and the square should add the thickness, ie:
```
square(is_list(gap) ? [gap.x, gap.y + thickness + eps] : [gap, thickness + eps], center = true);
```

nophead: Yes I think that is the way to go.

martinbudden:
I've fixed the belt length, so this should be ready to go.

PR #95, opened 2020-11-14
Converted belts test to coreXY.

martinbudden:
I've converted the belts test to CoreXY. The full CoreXY is shown below. This is actually done by stacking two half-CoreXY belts - the unstacked version is probably better for the test code, but I've shown both for reference.

beltsCoreXY

Half CoreXY:

beltsHalfCoreXY


nophead:
I preferred the two belt version, so I enabled it, as half a core XY doesn't make a lot of sense.

PR #96, opened 2020-11-14
Added support for square radial fans.

martinbudden:
square_radial_fan

### DIFF
--- a/vitamins/fan.scad
+++ b/vitamins/fan.scad
@@ -41,6 +41,9 @@ function fan_outer_diameter(type) = type[7];    //! Outside diameter of the fram
 function fan_blades(type)         = type[8];    //! The number of blades
 function fan_boss_d(type)         = type[9];    //! Diameter of the screw bosses
 function fan_aperture(type)       = type[10] ? type[10] : fan_bore(type); //! Optional diameter for the aperture, which can be bigger than the bore it has flared corners.
+function fan_offset(type)         = type[11];   //! Offset of fan from centre
+function fan_corner_inset(type)   = type[12];   //! Corner inset for screws
+function fan_base_height(type)    = type[13];   //! Height of base containing screws
 
 module fan(type) { //! Draw specified fan, origin in the centre
     width = fan_width(type);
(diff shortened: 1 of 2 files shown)


nophead:
Fan guard and mounting screws don't work with this because the mounting holes are at the bottom instead of the top. I.e. the screws are offset by the panel thickness plus the guard thickness from the base thickness, rather than the full height of the fan. Also the fan_guard needs to take the offset into account.

Probably it wouldn't need a guard as the blades are too short and close to trap a finger.

nophead:
Actually I think it is fairly easy to fix, so I will do it.

martinbudden:
I've adjusted the offset for the mounting screws:
radialFanWithBolts

The fan guard doesn't work, but these fans never seem to have guards, just like blower fans.

nophead:
The main problem is axial fans are panel mounted, so the screws go through the panel and ```fan_holes()``` makes a cut out and fan_guard can make either a guard or a grill, like in power supplies.

image

These functions could all be modified to take the offset into account and the screws go from above but does it make sense, would it ever be panel mounted like that?


This is actually a radial fan, which I have called a blower.

image


In particular it is screwed through its base to something underneath and it has a rectangular exit port. So logically it is one of these even if it has a different case shape.

martinbudden:
A radial fan would never be mounted like that, since it would be blowing the air in the wrong direction, so I don't think there is any point modifying those functions.

I did think of making the square radial fan a type of blower, but the blower code doesn't suit this.

I've only seen these square radial fans used on hotends, eg:

image

nophead:
Yes they are good for part cooling and perhaps air assist on a laser. That
is what I use radial blowers for but I need the exit port details to make
mating ducts.

It is unfortunate the shape is more like an axial fan but it isn't one.

nophead:
For example I use one here:

image

I would expect to be able to change the type to your square one and everything would adapt but it won't if it is modelled as an axial fan.

martinbudden:
If you like, I could try and redo it as a blower type.

nophead:
Yes please. I agree the code to draw it will get more ugly but the interface should be more correct, which is what matters to the end user.

martinbudden:
OK, will do.

martinbudden:
Closing. Superseded by https://github.com/nophead/NopSCADlib/pull/98

PR #98, opened 2020-11-15
Initial commit of square blower.

martinbudden:
I've also added fans to the old blowers.

Note that the duct is centralised. I don't know if your duct mating code assumes the duct is on the left, if so we may need a new duct_offset function.

blowers

### DIFF
--- a/vitamins/blower.scad
+++ b/vitamins/blower.scad
@@ -22,6 +22,7 @@
 //
 include <../utils/core/core.scad>
 use <../utils/rounded_cylinder.scad>
+use <screw.scad>
 
 function blower_length(type)      = type[2]; //! Length of enclosing rectangle
 function blower_width(type)       = type[3]; //! Width of enclosing rectangle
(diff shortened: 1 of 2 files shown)


nophead:
I have had to make a few changes to make it compatible using my laser fan assist as a test. I.e. I should just be able to change the blower type and everything should adjust.

As an aside ```type[0]``` should always match the name of the constant to make the documentation correct. I.e.
```blower(BL40x10)```.

```blower_lug(type)``` needs to be equal to the base thickness to place the screws correctly. It can't be added as a z offset in ```blower_screw_holes(type)``` because the XY positions without Z offset are needed for making holes and placing inserts, etc.

That means a zero lug height can no longer be used to decide it is square. I check the name begins with "Square" but it might need to be a type field if we get more types.

I had to add a ```blower_exit_offset(type)``` to be able to position a mating duct.

The ```blower_exit(type)``` function is supposed to be the outside dimension of the exit port so I added twice the wall width and removed your subtraction of  ```blower_wall(type)``` when calculating the corner_inset. Not sure why you subtracted one wall width, shouldn't it have been two if you thought it was the inside dimension?
```
function blower_casing_is_square(type) = slice(type[1], end = 6) == "Square";
function blower_exit_offset(type) = blower_casing_is_square(type) ? blower_length(type) / 2 : blower_exit(type) / 2; //! Offset of exit centre from the edge
```

Can you confirm the outside of the exit port is 30.6mm? That makes the corners too small for an M2 washer.

image

martinbudden:
To confirm measurements: external width of the exit port is 30mm precisely. There is room for M2 washer. Internal dimensions are 27.8mm by 8.5, so I think the wall width needs adjustment. There is flex in the top of the exit port, so an interference fit in the vertical direction is possible.

The subtraction of just one wall width was an error on my part.

nophead:
OK. The reason the washers didn't fit is that the corners are not pure arcs, they have a straight section and a smaller radius.

image

PR #104, opened 2020-12-13
Added sheet with checkerboard texture to simulate carbon fiber.

martinbudden:
When rendered, it looses its texture.

### DIFF
--- a/tests/sheets.scad
+++ b/tests/sheets.scad
@@ -27,5 +27,25 @@ module sheets()
     layout([for(s = sheets) width], 5)
         render_sheet(sheets[$i]) sheet(sheets[$i], width, width, 2);
 
+module test2() {
+    render_sheet(CF3) sheet(CF3, width, width, 2);
+    translate([40,0,0])
+        render_2D_sheet(CF3) sheet_2D(CF3, width, width, 2);
+    translate([80,0,0])
+        sheet(CF3, width, width, 2);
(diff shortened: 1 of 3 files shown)


nophead:
The way I model sheets is to do it all in 2D with sheet_2D and then use render_2D_sheet() unless it has 3D features like countersinks and then I use sheet() and render_sheet().

I am thinking render_2D_sheet() could be made to work with the woven pattern by doing 2D intersections with the weave and then linear_extruding each of them and then colouring. 

woven_sheet() would take a 2D child of the shape and then intersect it with one checker-board, linear_extrude and colour it and then do the same for the inverse checker-board. That would make it much simpler as it just needs enough squares, any excess gets trimmed by the intersection to take care of the remainder bits at the edge. You then don't need fillet to trim the corners, you can just pass a 2D sheet, which has the advantage they will be the correct colour. Currently they are green.

martinbudden:
Yes, that would make the edge handling easier. I'll look into it.

nophead:
I think render_sheet() could be made to sort of work by using projection() to sample just below the top and bottom surface and making two thin textured veneers. The edges and internal details would be one solid colour though.

martinbudden:
@nophead , can you have a look at this now, to see what you think?

martinbudden:
I think this is working fully now.

PR #108, opened 2020-12-16
Add optional length parameter to screw_polysink

martinbudden:
Added an optional length parameter to `screw_polysink`. This allows users to restrict the length of the cutout, primarily to avoid too many artifacts in the "Thrown Together" view in OpenSCAD.


nophead:
I just pushed a new version that has an h parameter. It is also wrapped in render() which should fix the artefacts, I think.

nophead:
Superseded by 157ff60e19310488cee6a15fb3bdc1df85811dfd

PR #109, opened 2020-12-17
Added rotating polycylinder.

martinbudden:
Here's the rotating polyhole. The image shows the rotating polyhole and the corresponding polyhole at a radius of 'M3_tap_radius'. The second column are the polyholes cut out of a cuboid, and the third column shows the top view of the cuboid. From above, the rotating polyhole looks almost perfectly round.

rotating_polyholes


martinbudden:
Closing, since superseded.

PR #110, opened 2020-12-20
Added render_if module.

martinbudden:
Added `render_if` module. Useful for `render_if(!$preview)` amoung other things.

### DIFF
--- a/utils/core/global.scad
+++ b/utils/core/global.scad
@@ -52,6 +52,13 @@ function slice(list, start = 0, end = undef) = let( //! Slice a list or string w
     ) is_string(list) ? slice_str(list, start, end) : [for(i = [start : 1 : end - 1]) list[i]];
 
 
+module render_if(render = true, convexity = 2)      //! Renders an object if ```render``` is true, otherwise leaves it unrendered
+    if (render)
+        render(convexity = convexity)
+            children();
+    else
+        children();
+
 module extrude_if(h, center = true)                 //! Extrudes 2D object to 3D when ```h``` is nonzero, otherwise leaves it 2D
     if(h)
         linear_extrude(h, center = center, convexity = 2) // 3D


nophead:
Why would you render only when $preview was false? Everything is calculated with CGAL when not not previewing.

martinbudden:
I have some complicated parts (funny shapes with lots of bolt holes) that don't slice properly the component parts aren't rendered before they are put together. The effect is very odd - the slicer gets confused and sort of turns the part inside out. So the voids become the part and the part becomes a void.

Strategic rendering of parts before they are `union`ed seems to solve the problem. However they look fine in F6, so there is no need to render for F6.

So doing this effectively "hints" the CGAL calculation.

nophead:
My understanding of render is as follows: Under F5 it forces geometry to be calculated with CGAL and cached. Under F6 if looks for cached geometry and uses it instead of calculating it, otherwise it does nothing and just calculate the geometry. So if you flush the cache before F6 it does nothing at all.

If you don't call the render during F5, I don't think it will do anything at all during F6 because it won't find a cached version. Does the actually do anything for you?

martinbudden:
I have a problem with rendering a complex shape that I haven't got to the bottom of. Judicial rendering seems to solve the problem, but I'm not entirely sure why. My object is now slicing properly, so I've put the investigation onto the back burner and will look into it some more once I've completed some other tasks.

nophead:
With this test;
```
$fn = 32;

render() difference() {
    sphere();
    cube();
}
```

Flush cache F6 gives this:

image
```
   Top level object is a 3D object:
   Simple:        yes
   Vertices:      467
   Halfedges:    2572
   Edges:        1286
   Halffacets:   1642
   Facets:        821
   Volumes:         2
```


F5 and then F6 gives this:

image

```
   Top level object is a 3D object:
   Facets:        930
```

So if you generate the STL from the GUI then render() can affect the result if you do F5 before F6 but if the STL is generated from the command line I don't think render() will have any effect at all. I could be wrong though as I vaguely remember adding them myself to get rid of CGAL errors but that might always have been in the GUI. Sometimes adding render() breaks things because it caches results as a polyset, which is lower resolution than CGAL. So then I have to flush the cache before F6 to a valid result.

PR #111, opened 2020-12-20
Added rounded_rectangles in the XZ and YZ planes.

martinbudden:
Add `rounded_rectangle_xz` and `rounded_rectangle_yz` modules that round in the respective planes.

Actually I think they should be called `rounded_cube_xz` and `rounded_cube_yz` and we should add `rounded_cube_xy`, while retaining the original `rounded_rectangle` module so as not to break old code.

I can see the original argument for not calling `rounded_rectangle` `rounded_cube` since it was not rounded on all 3 axes, but when there are three variants that argument no longer holds up.

### DIFF
--- a/tests/rounded_rectangle.scad
+++ b/tests/rounded_rectangle.scad
@@ -25,6 +25,12 @@ module rounded_rectangles() {
 
     translate([40, 0])
         rounded_rectangle([30, 20, 10], 3);
+
+    translate([80, 0])
+        rounded_rectangle_xz([30, 20, 10], 3);
+
+    translate([120, 0])
+        rounded_rectangle_yz([30, 20, 10], 3);
 }
 
 rounded_rectangles();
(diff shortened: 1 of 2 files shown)


nophead:
What do you use these for? Seems odd to just have rotated versions of something unless you use them very frequently. 

Yes the name has always been a bit of a niggle. Wouldn't a rounded cube have all eight corners rounded, or perhaps an option to select how it is rounded?

martinbudden:
I avoid square corners in `stl` files wherever I can. This helps avoid blobs or overextrusion on the corners and reduces ringing. Normally `rounded_rectangle` is fine. But then sometimes I rotate the `_stl` module to avoid overhangs, or reduce or avoid bridging etc.

Having these variants is very handy, since if halfway through designing a part I decide it should be oriented a different way for printing, I can just swap `rounded_rectangle` for (say) `rounded_rectangle_yz`.

nophead:
Yes that is what I mainly use it for but I always design my printer parts in the orientation they print in, which is why there was only the XY version and no rounded cube as it wouldn't be printable.

martinbudden:
> I always design my printed parts in the orientation they print in

So do I. But on occasion (normally when I putting the part into a bigger assembly) I realise I have the orientation wrong. Changing the orientation half way through is made much easier by having these variants.

PR #113, opened 2020-12-22
Added dogbones with circles offset in x and y directions.

martinbudden:
Added dogbones with the end circles offset in the x and the y directions. Useful if you need a dogbone near the edge of a part.

### DIFF
--- a/tests/dogbones.scad
+++ b/tests/dogbones.scad
@@ -26,6 +26,12 @@ module dogbones() {
     #translate([15, 0])
         dogbone_rectangle([10, 20, 5], center = false);
 
+    #translate([30, 0])
+        dogbone_rectangle_x([10, 20, 5], center = false);
+
+    #translate([45, 0])
+        dogbone_rectangle_y([10, 20, 5], center = false);
+
     sq = 3;
     translate([-5 + sq / 2 + eps, -10 + sq / 2 + eps])
         %cube([sq, sq, 1], center = true);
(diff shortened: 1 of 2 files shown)


nophead:
Rather than adding four new modules why not just add two parameters x_offset and y_offset defaulted to undef?

Then do x_offset = is_undef(x_offset) ? r / sqrt(2) : x_offset;

martinbudden:
Good idea. I'll alter the PR.

nophead:
I was thinking the same for dogbone_rectangle(). Just add x_offset and y_offset defaulted to undef and pass them to dogbone_square().

martinbudden:
I think `dogbone_rectange_x` and `dogbone_rectangle_y` are useful wrapper functions, it saves having to think about how much the offset is, and in which direction for the dogbone you want.

martinbudden:
OK, I've made the wrapper functions as thin as possible.

nophead:
I don't think I would use the wrappers as it would make a shape that needs filing. However I can see cases where a smaller offset can be used. If the hole is a clearance hole, they could be brought in to just touch at the corners and if the object that fits in the hole has a corner radius they could be brought in some more.

nophead:
Actually I see the point now after looking at the images.

martinbudden:
Yes, they are especially useful for doing the tabs on a sheet if you want to construct a box out of interlocking sheets, ie to do what woodworkers call finger joints.

martinbudden:
Here's and example of a finger joint using the new dogbones:

dogbones

nophead:
Nice, we should probably add pictures like to the blurb.

martinbudden:
Once we've got the carbon fiber sheets in, we can add some pictures using those.

PR #117, opened 2020-12-23
Martinbudden woven sheet

nophead:
I reimplemented this in a much simpler way because I had speed concerns with the original.


nophead:
Your version made the blanks woven, which I think would make operations on them very slow. I.e. drilling holes, etc.

I restored the original `sheet()` and `sheet_2D()` that are not coloured. 

When `render_2D_sheet()` is called it intersects the child with the chequerboard and extrudes it in each colour.

If `render_sheet()` is called I project the top and bottom surfaces and then 2D intersect with chequerboard and add two thin veneers. Projection is slow but I think it will be a lot faster than doing 3D operations on a 3D chequerboard.

The downsides are the 3D version is colour2 on the inside and edges but the 2D versions is chequerboard all the way through. Since the sides and holes are normally covered in the assembly I don't think it will matter.

image

image

The other downside is the size needs passing to `render_sheet()` and `render_2D_sheet()` when the sheet is woven as they need to know the size of the chequerboard.

martinbudden:
Yes, that's much better. Mine started to slow down considerably once you had about 400 squares in the checkerboard.

It turns out this is a useful debugging tool, by setting the weave to (say) 5, you can easily spot gross errors in your positioning of cutouts on the sheet.

One minor quibble: the dimensions of the sheet are specified as `w` and `d`, presumably for width and depth. Can't you just use `x` and `y`? Depth is often associated with the z axis, and width can be in the y direction when associated with a length in the x direction. `x` and `y` is unambiguous.

CF_dogbones

nophead:
They are w and d to match sheet() and sheet_2D() parameters that they have to mirror.

If I was starting again I would use size everywhere and .x, .y and .z.

I seem to have found a bug in OpenSCAD as I sometimes see black line but it disappears when I pan or rotate.

image

PR #121, opened 2020-12-29
Added GT2x16_pulley.

martinbudden:


nophead:
I changed the name to GT2 as "16 teeth" gets added to the description.  Where there are different hub styles I have suffixes to distinguish them. I have added comments to show they stand for UltiMaker, OpenBuilds and RobotDigg.

PR #122, opened 2020-12-31
Added size and slot convenience functions. Added non-rounded hole.

martinbudden:
Non-rounded hole is for when you need to bridge around the rocker hole.


nophead:
Why not rounded? The slot is longer than the switch and the tabs only make contact away from the corners.

martinbudden:
The non-rounded form is for when you need to bridge the hole around the rocker switch, for example if you are putting it into a vertically printed panel.

nophead:
I normally use horizontal = true to indicate that. If the radius was important it would turn it into a tearslot but in this case it isn't.

martinbudden:
I thought about using `horizontal = true`, but it seemed confusing for the rocker switch. For a screw, the meaning is clear, since symmetry means there is only one horizontal orientation. However for the switch, there are two ways to rotate it that could both conceivably mean horizontal only one of which requires a non-rounded hole, so I opted for a direct description of what the parameter was for.

PR #124, opened 2021-01-08
Conditionally rendered pulleys to speed up drawing.

martinbudden:
Also changed "silver" to silver.

### DIFF
--- a/vitamins/pulley.scad
+++ b/vitamins/pulley.scad
@@ -108,24 +108,26 @@ module pulley(type) { //! Draw a pulley
                         cylinder(r = screw_radius(pulley_screw(type)), h = 100);
     }
 
-    color("silver") {
-        render() difference() {
-            rotate_extrude() translate([r1, 0]) {
-                if(hl)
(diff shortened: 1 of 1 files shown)


nophead:
LOL, I am currently trying to find missing renders to speed up drawing! Do you mean to speed up the rendering()?

martinbudden:
Yes, that's what I meant.

nophead:
Having tried both versions a few times I don't see any consistent improvement in the test times.

I think that is because when the difference has only one operand it gets short circuited and then the render only works on the linear extrude of the teeth, which is just a polyhedron already, so render takes no time in that case.

nophead:
Note also some release candidates made pulleys a lot slower, see https://github.com/openscad/openscad/issues/3529

nophead:
OK I merged a version that should be optimum on all versions. It always renders the difference but avoids the difference altogether when there isn't a hole.

martinbudden:
OK, thanks for checking into this.

PR #126, opened 2021-01-09
Changed linear rails to use a v-slot rather than a square slot.

martinbudden:
This looks a lot better, and the slot is more visible.

It also means that linear rails are now printable, and even usable

There are a number of reasons you might want to print a linear rail

1. to build a prototype while waiting for parts to arrive
2. for a cheap or not often used project where high precision is not required
3. to follow the original RepRap philosophy of maximising printed parts

(Note that currently the dimensions have not been tuned or tested for use as printed parts, however they are fine for just drawing the parts. The video shows a MGN12 carriage being used on a printed rail - there is still a bit of play but I am confident that can be reduced to make the printed rail genuinely useful in certain projects).

https://user-images.githubusercontent.com/194586/104091738-3d6be880-5277-11eb-99fd-861702f4caa7.MOV

### DIFF
--- a/vitamins/rail.scad
+++ b/vitamins/rail.scad
@@ -33,6 +33,9 @@ function rail_bore_depth(type) = type[7];   //! Counter bore depth
 function rail_screw(type)      = type[8];   //! Screw type
 function rail_carriage(type)   = type[9];   //! Carriage type
 function rail_end_screw(type)  = type[10];  //! Screw used for ends only (Countersink used for better location)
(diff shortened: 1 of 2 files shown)


nophead:
The smallest rail's groove does render correctly but I can't see why, the polygon looks OK to me.

image

nophead:
OK I found the problem and fixed it. The bottom core of the rail was the full with and cut off the groove.

PR #127, opened 2021-01-09
Added printable pulleys.

martinbudden:
The test code shows the printed pulleys along side the pulley vitamins.

The main differences are 
1. the printed pulleys have a 45 degree top flange
2. Teardrops and polycircles are used
3. Changes to the tooth profile for GT2 pulleys
  
PrintedPulleys

Note - this is a remix of droftarts's printed pulleys on thingiverse. I've credited him in the blurb as per the terms of the Creative Commons - Attribution - Share Alike license that he licensed it under. Funnily enough, he credits you in his blurb.

### DIFF
--- a/printed/printed_pulleys.scad
+++ b/printed/printed_pulleys.scad
@@ -1,0 +1,182 @@
+//
+// NopSCADlib Copyright Chris Palmer 2018
+// nop.head@gmail.com
+// hydraraptor.blogspot.com
+//
+// This file is part of NopSCADlib.
+//
+// NopSCADlib is free software: you can redistribute it and/or modify it under the terms of the
+// GNU General Public License as published by the Free Software Foundation, either version 3 of
+// the License, or (at your option) any later version.
+//
+// NopSCADlib is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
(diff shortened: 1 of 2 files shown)


nophead:
When I print pulleys I don't use a clearance hole. I have only done them from somebody else's STL, so the hole comes out too small and then I ream it to get a nice snug fit. When you ream soft plastic like ABS the result is slightly too small due to spring back but it makes an nice interference fit with absolutely no play.

martinbudden:
Good point. Should I just use the exact diameter? Ie 3 rather than 2*M3_clearance_radius?

I also thought of putting an optional flat piece in the timing pulleys, to mate with a NEMA motor's ground shaft.

nophead:
Yes certainly no clearance. I deliberately under size holes that I am going to ream. The pulleys I have reamed have an exact 5mm bore in the STL but they aren't polyholes, so they print undersized.

martinbudden:
Thinking about this a bit more: you certainly want a tight fit for timing pulleys. However for idler pulleys that may not be the case - you may want it to just spin on a bolt. I guess the user can drill it out if they don't want a tight fit though.

nophead:
Even for an idler I would want a round hole, not a polyhole, for smooth running. A reamed hole should spin on a bolt as bolt shafts are normally a bit smaller than nominal.

PR #133, opened 2021-01-13
Added ESP8266.

martinbudden:


nophead:
I think it needs to be rotated 90 degrees to match the others.

Also I think the name needs to be of the board, not the chip as I use ESP8266 ESP-12F that looks totally different.

martinbudden:
OK, I've renamed it to ESP-01 and rotated it.

PR #143, opened 2021-02-01
Made C version of linear rails and carriages explicit.

martinbudden:
Renamed `MGN7 ` etc to `MGN7C` to make the variant of the rail and carriage explicit.

Without this rename you have to find a linear rail dimension chart to work out which version of the rail/carriage you are getting.

### DIFF
--- a/tests/rails.scad
+++ b/tests/rails.scad
@@ -26,15 +26,16 @@ sheet = 3;
 pos = 1; //[-1 : 0.1 : 1]
 
 module rails()
-    layout([for(l = rails) carriage_width(rail_carriage(l))], 20)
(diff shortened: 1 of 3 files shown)


nophead:
Hmm, looks like the carriage should have an association to its rail rather than the other way around. Then if we have two carriage types for the same rail we don't have to duplicate the rail data, more correct but more breaking. I.e.

Remove rail_carriage() and add carriage_rail(), replace rail_travel() with carriage_travel().
No need to pass rail to carriage().
rail_assembly() would take a carriage instead of a rail.

martinbudden:
OK, I'll update the PR along those lines.

martinbudden:
Done. It looks much better that way.

PR #144, opened 2021-02-03
Corrected MGN9 rail end value.

martinbudden:
`rail_end` value for MGN9 rails is 5mm, not 7.5mm.

### DIFF
--- a/vitamins/rails.scad
+++ b/vitamins/rails.scad
@@ -23,7 +23,7 @@
 //                Wr  Hr    E     P   D    d    h                                       go    gw
 MGN5 = [ "MGN5",   5,  3.6,  5,   15, 3.6, 2.4, 0.8, M2_cs_cap_screw, M2_cs_cap_screw,  1,    1 ]; // Screw holes too small for M2 heads
 MGN7 = [ "MGN7",   7,  5,    5,   15, 4.3, 2.4, 2.6, M2_cap_screw,    M2_cs_cap_screw,  1.5,  1.5 ];
-MGN9 = [ "MGN9",   9,  6,    7.5, 20, 6.0, 3.5, 3.5, M3_cap_screw,    M3_cs_cap_screw,  1.5,  1.5 ];
+MGN9 = [ "MGN9",   9,  6,    5,   20, 6.0, 3.5, 3.5, M3_cap_screw,    M3_cs_cap_screw,  1.5,  1.5 ];
 MGN12 =[ "MGN12", 12,  8,   10,   25, 6.0, 3.5, 4.5, M3_cap_screw,    M3_cs_cap_screw,  2.25, 2.75];
 MGN15 =[ "MGN15", 15, 10,   10,   40, 6.0, 3.5, 5.0, M3_cap_screw,    M3_cs_cap_screw,  2.5,  2.75 ];
 SSR15= [ "SSR15", 15, 12.5, 10,   60, 7.5, 4.5, 5.3, M4_cap_screw,    M4_cs_cap_screw,  2.5,  2.75 ];


nophead:
Seems to depend where you get them from. Here it 7.5mm. https://www.aliexpress.com/i/32808719639.html and here where I bought min from: https://robotdigg.com/product/779/Quality-440C-SUS-MGN9-linear-rail-with-carriage

It changes the number of holes in the machine I am working on at the moment, so I think I have to revert this. Unfortunately I can't check my rails at the moment.

martinbudden:
Actually, how far the first hole if offset is dependent on the rail length rather than the manufacturer. So I have a 150mm MGN9 rail with a first hole offset of 5mm, and a 200mm MGN9 rail with a first hole offset of 10mm, they were ordered from the same supplier.

I have a function to deal with this.
```
function railFirstHoleOffset(type, length) = (length - (rail_holes(type, length) - 1)*rail_pitch(type))/2;
```

nophead:
It isn't the distance to the first hole, it is the minimum that it can be so it determines the number of holes given a rail length. 

I work out the number of holes by subtracting twice its value from the length and dividing by pitch. The holes are then centred and that determines the actual firstHoleOffset, which will be as your formula if the number of holes is correct.

The problem is the number of holes has gone up by one on my 250mm rails.

nophead:
Reading up on the, the rails are made in long lengths with all the holes and then cut to size with an equal gap at each end, so perhaps there is a more complex rule in play as there is also a maximum E before you hit the next hole.

nophead:
Thinking about this.

There are two possible hole patterns: an odd number with a hole in centre or an even number with the centre halfway between two holes. If you consider the odd case with say three holes then the shortest length of half of it is `pitch + E`. When the half length is infinitesimally less it switches to two holes and the distance to the first hole is then `pitch + E - pitch / 2` = `E + pitch / 2`.

In order not to break into the next hole `E + pitch / 2 < pitch - D / 2`. Where D is the counter-bore diameter. So E must be less than `(pitch - D) / 2`. So in the case of MGN9 it has to be less than 7mm, so 7.5mm doesn't make sense unless they manufacture to a specific length and not cut them from a longer rail, or they don't offer some lengths. Some of the other rails violate this rule as well, so I am confused.

To validate this theory I changed the code to subtract two more holes, one past each end. Setting the length to some values then gets a rail with dimples in the end, showing it is impossible to cut from a larger rail without using a smaller E value so that it switches to one more hole earlier.

I think I will have to wait until I am at home and check my rails. We might need an extra parameter to some of the rail functions if my 250mm rail has a different end distance to your 150mm rail as both should be the same because the difference in length is a multiple of the pitch.

nophead:
So I counted the holes in my MGN9 rails and they are correct with this change. 250mm had one hole less with the datasheet value.

I think the issue is the E value on the datasheet is the value they use when they sell them in preferred lengths, which are just a multiple of pitch + 2 E. 

`rail_end()` is actually Emin, which they say E has to no less than but don't state it anywhere I can find. I have updated the column name in `rails.scad` to make it clear it is Emin, rather than E.

I added an assert for values of Emin too big and found the value for MGN12 was too big, so I reduced it to 7.5mm.

I don't know what the correct value for Emin is. It could potentially be as small as width / 2 without compromising the screw holes. Making it bigger than that means the maximum end length gets bigger and potentially means more of the rail not secured. I.e. when the ends are bigger than pitch / 2, which they have to be for some lengths. However I think may be they just don't sell it in lengths that don't have sensible ends.

All I am using Emin for is to calculate the number of holes a given length will have. If only a subset of possible lengths are sold then it just needs to be small enough to get those hole counts right. My guess would be that using width / 2 would probably be correct because given the choice of coming as close to the screw hole as the sides do would be preferable to one screw less and having width / 2 + pitch / 2 at the ends. However, I don't think such lengths are available so we will probably never know.

PR #145, opened 2021-02-03
Rounded rectangles revisited

martinbudden:
Having used the new rounded rectangles for a while, I think they should be renamed to `rounded_cubes` and adopt the default centering of the cube.

1. They are 3D objects, so `rounded_cube` is a more accurate name than `rounded_rectangle`
2. The argument that they should not be called a `rounded_cube` because they are only rounded in one plane doesn't apply because of the rounding plane suffix, eg `rounded_cube_xy`
3. Once you adopt the `rounded_cube` name, then the default centering should be the same as for a `cube`.

I've left the original `rounded_rectangle` to keep compatibility with old code.

### DIFF
--- a/tests/rounded_rectangle.scad
+++ b/tests/rounded_rectangle.scad
@@ -27,10 +27,13 @@ module rounded_rectangles() {
         rounded_rectangle([30, 20, 10], 3);
 
     translate([80, 0])
-        rounded_rectangle_xz([30, 20, 10], 3);
+        rounded_cube_xy([30, 20, 10], 3);
 
     translate([120, 0])
-        rounded_rectangle_yz([30, 20, 10], 3);
+        rounded_cube_xz([30, 20, 10], 3);
+
+    translate([160, 0])
+        rounded_cube_yz([30, 20, 10], 3);
 }
 
 rounded_rectangles();
--- a/utils/core/rounded_rectangle.scad
+++ b/utils/core/rounded_rectangle.scad
@@ -22,32 +22,33 @@
 //
 module rounded_square(size, r, center = true) //! Like `square()` but with with rounded corners
 {
+    assert(r < min(size.x, size.y) / 2);
     $fn = r2sides4n(r);
     offset(r) offset(-r) square(size, center = center);
 }
 
 module rounded_rectangle(size, r, center = true, xy_center = true) //! Like `cube()` but corners rounded in XY plane and separate centre options for xy and z.
 {
-    linear_extrude(size.z, center = center)
+    extrude_if(size.z, center = center)
         rounded_square([size.x, size.y], r, xy_center);
 }
 
-module rounded_cube_xy(size, r, center = false, xy_center = false) //! Like `cube()` but corners rounded in XY plane and separate centre options for xy and z.
+module rounded_cube_xy(size, r = 0, xy_center = false, z_center = false) //! Like `cube()` but corners rounded in XY plane and separate centre options for xy and z.
 {
-    linear_extrude(size.z, center = center)
+    extrude_if(size.z, center = z_center)
         rounded_square([size.x, size.y], r, xy_center);
 }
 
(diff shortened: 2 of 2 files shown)


nophead:
Not sure I like the new name any more than the old. How about cube_rounded_in_xy() to be more descriptive and unambiguous?

martinbudden:
Generally I prefer using a suffix rather than a prefix as a qualifier, so generally I'd prefer `cube_rounded` to `rounded_cube`. I don't generally like to combine a prefix or a suffix with a preposition though, so I'd prefer `cube_rounded_xy()` to `cube_rounded_in_xy()`.

The reason I initially proposed `rounded_cube` is because there are already quite a few `rounded_` objects in the library, and generally I like to maintain consistency if possible.

martinbudden:
Just to clarify my previous comment. If this was being written from scratch, I'd prefer the name `cube_rounded_xy`, but since there are already `rounded_*` modules in the library, I prefer `rounded_cube_xy`.

nophead:
Only in maths is a rectangle 2D, in real life rectangles have a thickness so I would call a flat sheet a rectangle with rounded corners, not a cube rounded xy.

martinbudden:
There are at least 4 different things to consider when choosing this name:

1. Common parlance
2. Established OpenSCAD convention
3. Mathematical correctness
4. Programming usage patterns.

1. Dealing with common parlance first. Agreed, I would also call a flat sheet a rectangle. But whether a cuboid is called a rectangle depends somewhat on its dimensions. A [100,50,5] cuboid would probably be called a rectangle. But a [50,50,40] cuboid would probably be called a cube. And a [300,20,20] cuboid would probably be called a bar. And a [100,80,40] cuboid might be called a box or a block. So common parlance doesn't really give a clear resolution as to what the name should be.

2. OpenSCAD convention is very clear. 2D shapes use a 2D name (square, circle) and 3D shapes use a 3D name (cube, sphere, cylinder) and the symmetric name is used even if the object is not symmetric (ie square rather than rectangle, cube rather than cuboid).

3. Neither rectangle nor cube are mathematically correct.

4. A usage pattern I fairly regularly use, and I imagine others do too, is to build but my object out of simple shapes (eg cubes, cylinders etc) and then round them as appropriate, so changing the cube to (currently) a rounded_rectangle. Currently this is a real pain, because I have to change the name and add the centering parameters. Now if the object is called rounded_rectangle_xy, then it should have the same centering defaults a rounded rectangle, but if you want the same centering defaults as a cube, then it should be called rounded_cube_xy.

For me (2) and (4) are fairly conclusive arguments. But then I seem to have a higher desire than you do to adhere to OpenSCAD conventions.

nophead:
I don't find the centering option of cube useful. I rarely need it centred in all axes but I often want it centred in xy but not z.

martinbudden:
> I don't find the centering option of cube useful. I rarely need it centred in all axes but I often want it centred in xy but not z.

Exactly. A fully centered cube is not really very useful. And `rounded_rectangle` defaults to fully centered - which is what I want to change.

nophead:
Yes centre should have defaulted to false and xy_center to true as that is by far the most common use. If xy_centre defaulted to false like cube it would be just as useless.

martinbudden:
I've just made two more changes (leaving aside the naming debate for the moment).

1. Added an assertion in `rounded_square` to check `r` is not too large. If `r` is too large, nothing is displayed and this can be confusing.
2. Changed the rectangles to use `extrude_if` so the rectangles collapse squares if `size.z` is zero.

(2) may seem a bit odd, but I have an actual use case for it. In my current project I am providing alternative builds - either a fully 3D printed part, or a carbon fiber sheet with printed parts attached. The cutouts in the carbon fiber sheet have to match the cutouts in the 3D printed part, so when I `difference` my `sheet_2D` I want to use the same cutouts, but just specify `size.z = 0`.

PR #149, opened 2021-02-08
Included core.scad in polyholes.scad

martinbudden:
Included `core.scad` in polyholes, so polyholes can be `use`d rather than `include`d.


nophead:
core.scad uses global.scad that includes polyholes.scad, so you never need to include it. It is used when you include core.scad.

martinbudden:
I should have more thoroughly checked...

PR #150, opened 2021-02-09
Allow poly_shapes to collapse to non-poly form.

martinbudden:
Changed the default value of `sides` in `poly_circle` to `1` (which was a previously illegal value) and use this to invoke the calculation of `sides(r)`. This leaves the value of `0` free so that `circle` can be called with `$fn=0`, so the `poly_circle` collapses into a circle.

So `poly_circle(r)` has no change in behaviour, and `poly_circle(r, sides=0)` gives a circle.

`poly_cylinder` and `poly_ring` are similarly changed.

The motivation behind this is to make it easier to use the same code to generate printed and CNC'ed parts.

### DIFF
--- a/utils/core/polyholes.scad
+++ b/utils/core/polyholes.scad
@@ -33,20 +33,20 @@
 //! When `twist` is set the resulting cylinder is extended by `eps` at each end so that the exact length of the hole can be used without
 //! leaving a scar on either surface.
 //
-function sides(r) = max(round(4 * r), 3);                                       //! Optimium number of sides for specified radius
-function corrected_radius(r, n = 0)   = r / cos(180 / (n ? n : sides(r)));      //! Adjusted radius to make flats lie on the circle
-function corrected_diameter(d, n = 0) = d / cos(180 / (n ? n : sides(d / 2)));  //! Adjusted diameter to make flats lie on the circle
+function sides(r, n = undef) = is_undef(n) ? max(round(4 * r), 3) : n ? max(n, 3) : r2sides(r); //! Optimium number of sides for specified radius
+function corrected_radius(r, n = undef)   = r / cos(180 / sides(r, n));                         //! Adjusted radius to make flats lie on the circle
+function corrected_diameter(d, n = undef) = 2 * corrected_radius(d / 2 , n);                    //! Adjusted diameter to make flats lie on the circle
 
-module poly_circle(r, sides = 0) { //! Make a circle adjusted to print the correct size
-    n = sides ? sides : sides(r);
-    circle(r = corrected_radius(r,n), $fn = n);
+module poly_circle(r, sides = undef) { //! Make a circle adjusted to print the correct size
+    n = sides(r, sides);
+    circle(r = corrected_radius(r, n), $fn = n);
 }
 
-module poly_cylinder(r, h, center = false, sides = 0, chamfer = false, twist = 0) {//! Make a cylinder adjusted to print the correct size
+module poly_cylinder(r, h, center = false, sides = undef, chamfer = false, twist = 0) {//! Make a cylinder adjusted to print the correct size
     if(twist) {
         slices = ceil(h / layer_height);
         twists = min(twist + 1, slices);
-        sides = sides ? sides : sides(r);
+        sides = sides(r, sides);
         rot = 360 / sides / twists * (twists < slices ? (1 + 1 / slices) : 1);
         if(center)
             for(side = [0, 1])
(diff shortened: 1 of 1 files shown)


utils/core/polyholes.scad
```diff
@@ -34,15 +34,15 @@
 //! leaving a scar on either surface.
 //
 function sides(r) = max(round(4 * r), 3);                                       //! Optimium number of sides for specified radius
-function corrected_radius(r, n = 0)   = r / cos(180 / (n ? n : sides(r)));      //! Adjusted radius to make flats lie on the circle
-function corrected_diameter(d, n = 0) = d / cos(180 / (n ? n : sides(d / 2)));  //! Adjusted diameter to make flats lie on the circle
+function corrected_radius(r, n = 0)   = r / cos(180 / (n == 0 ? r2sides(r) : n == 1 ? sides(r) : n));    //! Adjusted radius to make flats lie on the circle
```
nophead, line 37: Default n should be 1?

utils/core/polyholes.scad
```diff
@@ -34,15 +34,15 @@
...
+function corrected_diameter(d, n = 0) = corrected_radius(d / 2, n);                                     //! Adjusted diameter to make flats lie on the circle
```
nophead, line 38: Likewise

utils/core/polyholes.scad
```diff
@@ -34,15 +34,15 @@
...
 
-module poly_circle(r, sides = 0) { //! Make a circle adjusted to print the correct size
-    n = sides ? sides : sides(r);
-    circle(r = corrected_radius(r,n), $fn = n);
+module poly_circle(r, sides = 1) { //! Make a circle adjusted to print the correct size
+    n = sides == 1 ? sides(r) : sides;
```
nophead, line 41: Doesn't like correct to me. n should either be sides(r) or r2sides(r)

utils/core/polyholes.scad
nophead, line 38: Should be twice the corrected_radius()

nophead:
I use poly_drill() to do this.

I think it would be cleaner to use undef instead of 1 to represent use sides(r).

Perhaps add an optional n = undef parameter to sides() and handle it all there.

utils/core/polyholes.scad
martinbudden: This has a facility for the user to force the number of sides, ie `poly_circle(5, 8)` forces an octagon. However, I think I have overlooked the n=0 case. I think the correction should be:
```
n = sides == 0 ? r2sides(r) : sides == 1 ? sides(r) : sides;
```
Do you agree?

nophead: Yes that is correct but I think passing n to sides() and letting it sort out the number of sides will make the code a lot simpler. E.g.
```
function sides(r, n = undef) = is_undef(n) ? max(round(4 * r), 3) : n ? n : r2sides(r);                                       //! Optimium number of sides for specified radius or overridden by n which if 0 uses r2sides(r)
```

Then default `sides` to `undef` in all the clients and I think it will all get simpler. E.g.
```
function corrected_radius(r, n = undef)   = r / cos(180 / sides(r, n));      //! Adjusted radius to make flats lie on the circle

module poly_circle(r, sides = undef) { //! Make a circle adjusted to print the correct size
    n = sides(r, sides);
    circle(r = corrected_radius(r, n), $fn = n);
}
```
Etc.

martinbudden:
I've changed this to handle this with `undef` and rebased. Looks much better.

Yes, I've seen `poly_drill`, the trouble with that is it selects `circle` vs `poly_circle` on the global `cnc_bit_r`, which makes it fine for use across variants, but it does not work for the use case of having two versions of a single part, one printed and one CNC'ed, in a single variant.

nophead:
I just do something like this to switch a feature from cnc able to 3D printable.

```
                    pcb = opengrab_pcb();
                    let($cnc_bit_r = 0)
                        pcb_cutouts(pcb);
```

That gives me rectangular cutouts instead of dog_bone versions. Any round holes would switch to poly_holes().

utils/core/polyholes.scad
martinbudden: Do you need to explicitly assign `n =  undef``? Surely it is undefined if it is not assigned a value?

nophead: I think its clearer that you can omit it and possibly OpenSCAD will warn if you don't pass enough parameters in the future as it gets more picky all the time.

Why did you not take my suggestion to handle it all in sides()?

I think poly_circle() doesn't work with sides = 0 in the current iteration.

martinbudden: > Why did you not take my suggestion to handle it all in sides()?

I didn't see it. I'll make that change and resubmit.

martinbudden:
I've made the sides function change. I've also changed the diameter code back to my initial submission because I think that is correct. Not sure why you said the corrected diameter should be twice the corrected radius.

nophead:
Isn't a diameter always twice the radius? How can corrected_diameter = corrected_radius make sense?

martinbudden:
The code is `function corrected_diameter(d, n = 0) = corrected_radius(d / 2, n);` - note the `d / 2`

The old code was:
```
function corrected_radius(r, n = 0)   = r / cos(180 / (n ? n : sides(r)));      //! Adjusted radius to make flats lie on the circle
function corrected_diameter(d, n = 0) = d / cos(180 / (n ? n : sides(d / 2)));  //! Adjusted diameter to make flats lie on the circle
```

The same relationship applies.

nophead:
Well try comparing the results!

The old code divides d by the cosine. The new code just uses corrected_radius, which divides r by the same cosine.

martinbudden:
Ooops! I'll correct it.

nophead:
Poly_cylinder needs some changes, e.g. here https://github.com/nophead/NopSCADlib/blob/eff17cb38b221405017958d83f50e3cb559abd23/utils/core/polyholes.scad#L49

martinbudden:
Fixed and rebased. Should be OK now.

PR #160, opened 2021-03-06
Added parameterised CoreXY.

martinbudden:
This is a fully parameterised implementation of CoreXY. It supports virtually all variants of CoreXY that do not used crossed belts.

In particular the separation of idler pulleys on the Y carriages is parameterised, so these pulleys may be offset from each other, or be coincident in the X and/or Y directions.

Additionally the drive pulleys may be offset, to allow more flexible positioning of the motors.

The drive pulleys and idler pulleys may be different sizes.

The test code displays the belts and pulleys and adds screws to illustrate how pulleys and screws are positioned.

The test code is quite complicated because it allows all variants to be tested. An actual implementation would choose one variant and so would be considerably simpler.

core_xy_belts


nophead:
Please can you change the filenames to just coreXY. You already have the record for the longest names in several columns and that has broken the GitHub layout but I didn't notice until too late.

I think perhaps the files for KP_pillow_blocks could be pillow_blocks and SCS_bearing_blocks just bearing_blocks and rounded_right_triangle rounded_triangle to bring the width down. The code could keep the same names, so it is minimally breaking, just the include name when not using lib.scad.

martinbudden:
Renames. I'll put in PRs to rename the pillow blocks and bearing blocks as well.

nophead:
Why are the pulleys excluded the BOM?

martinbudden:
The pulleys are only there for debugging - to ensure you have your pulleys in the correct place. Any implementation is going to have the pulleys as parts of different assemblies. The idea is that you set `show_pulleys=true` to check your pulleys are in the correct position.

However, having said that, it's reasonable to have the pulleys on the bom if `show_pulleys=true`. I'll change it.

PR #161, opened 2021-03-06
Reordered openscad command line parameters to put output name first.

martinbudden:
Having the output file name first makes it easier to see what file the build is currently working on, which is handy for a long build.


nophead:
Good idea

PR #162, opened 2021-03-06
Improved resilience of compare_images.

martinbudden:
Sometimes `magick compare` returns a string, eg

"compare: image widths or heights differ `tests/png/frontface.png' @ error/compare.c/CompareImagesCommand/1160."

This PR makes `tests.py` resilient to that.


nophead:
Perhaps something in more recent versions of Magick as I haven't encountered it.

PR #168, opened 2021-03-11
Open belt

SmoothieAq:
Extension to belt.scad

Can now:
- render open loops
- twist the belt
- use pulleys instead of radius in the points list

Fixes some precision a few places
Breaking change in belt_length(); now requires a type argument

NOTE, this PR requires 2021.01 and also adds a few helper methods to maths.scad. If you
don't like that, then just say so, and I will change the code to not have the requirement.


nophead:
It will take me some time to get my head around this. A few points so far:

What do you mean by "Fixes some precision a few places"?

Why does the length now depend on the belt type? What was wrong with modelling the pitch line to avoid that?

We already have a slice() function with Python semantics here: https://github.com/nophead/NopSCADlib/blob/master/utils/core/global.scad#L48

I don't like using .x and .y for [0] and [1] when it isn't an X or Y coordinate as it makes the code harder to read (although I may have done that myself!).

https://github.com/nophead/NopSCADlib/pull/168/files#diff-89f5671faec9d71156b48f8f82f636295ecee006fb9fbc923f0a7b8d5a5b1566R105: why not just
```
mirrored = xor(twisted[i], p.z < 0) ? 0 : 1;
```

SmoothieAq:
> It will take me some time to get my head around this. A few points so far:
Sure, and no hard feelings a all if you just don't like it ;-)

> What do you mean by "Fixes some precision a few places"?
I'll look it up and get back on that.
> 
> Why does the length now depend on the belt type? What was wrong with modelling the pitch line to avoid that?
The reason is that you now can specify the path with pulleys instead of radius. The radius is then calculated as a combination of the pulley radius and the belt thickness - the belts I use have a different thickness than the ones that are "baked-in" in the pulley types. Also the twist influences the calculation of the radius.
> 
> We already have a slice() function with Python semantics here: https://github.com/nophead/NopSCADlib/blob/master/utils/core/global.scad#L48
Missed that, I will change that.
> 
> I don't like using .x and .y for [0] and [1] when it isn't an X or Y coordinate as it makes the code harder to read (although I may have done that myself!).
Yes, I don't disagree, I will try to change that.
> 
> https://github.com/nophead/NopSCADlib/pull/168/files#diff-89f5671faec9d71156b48f8f82f636295ecee006fb9fbc923f0a7b8d5a5b1566R105: why not just
> 
> ```
> mirrored = xor(twisted[i], p.z < 0) ? 0 : 1;
> ```
I'll check and get back on that...

nophead:
>The reason is that you now can specify the path with pulleys instead of radius. The radius is then calculated as a combination of the pulley radius and the belt thickness - the belts I use have a different thickness than the ones that are "baked-in" in the pulley types. Also the twist influences the calculation of the radius.

The pitch radius is a fixed property of the pulley type, which is why it is a property and it is independent of the belt thickness. when it is toothed. When the back of the belt goes around a smooth pulley then it depends on the belt thickness. So in that case pulley_pr() should take an optional belt parameter if you are using a different belt.

I assume the reinforcing cords are placed at the pitch radius and that is where the length is constant.

nophead:
I don't think we need two versions of `rounded_polygon_tangents()`. I think its only use is to pass them to `rounded_polygon()` and `rounded_polygon_length()`, so the new data format is no problem.

We can also have a `rounded_polygon()` function that returns a point list now you have added `rounded_polygon_arcs()`. I was planning to do that so it can be passed to `sweep()` so I can add it after this is merged.

Now you have added the open option the gap stuff is redundant. Since `belt_length()` taking a type is breaking I don't mind breaking the gap stuff if it is straight forward to use the open option.

I don't think `belt_length()` should round up to a multiple of teeth. I use it to alliteratively find pulley distances for closed belts and it wont converge properly with rounding. Just the number on the BOM is rounded.

nophead:
I updated pulley_pr() to take an optional belt type.

SmoothieAq:
> > The reason is that you now can specify the path with pulleys instead of radius. The radius is then calculated as a combination of the pulley radius and the belt thickness - the belts I use have a different thickness than the ones that are "baked-in" in the pulley types. Also the twist influences the calculation of the radius.
> 
> The pitch radius is a fixed property of the pulley type, which is why it is a property and it is independent of the belt thickness. when it is toothed. When the back of the belt goes around a smooth pulley then it depends on the belt thickness. So in that case pulley_pr() should take an optional belt parameter if you are using a different belt.
> 
> I assume the reinforcing cords are placed at the pitch radius and that is where the length is constant.

Yes, there is however also the case where you run the toothed side on a smooth pulley (an idler). If you review the different current coreXY designs, you will see that they fall in three groups: some use pulleyes as idlers (as you do in the tests/belts), some use the toothed side against a smooth pulley (e.g. https://www.thingiverse.com/thing:2254103), and some twist the belt to get the smooth side against a smooth pulley (https://www.blvprojects.com/blv-mgn-cube-3d-printer). To calculate the radius to use for the belt path if we want to support these three style, we will need to know both which belt side is against the pulley and the specific type of belt. This is the reason for my suggested belt_pulley_pr(type, pulley, twisted) method, although it could as well have been pulley_pr(type, belt, twisted).

SmoothieAq:
> What do you mean by "Fixes some precision a few places"?

In rounded_polygon_length(points, tangents) I believe the calculation of the length of an arc is not correct if `r` is negative - compare it with rounded_polygon_arcs(points, tangents). I discovered this because I actually use the arc calculations to extrude the belt.
The other case is where you choose to run the toothed side against a smooth pulley, this is not handled in the current code.

nophead:
OK I will have a look at that.

I never run toothed belts over smooth pulleys as it gives rise to lumpy motion. It was a problem on early RepRap machines that I fixed here: https://hydraraptor.blogspot.com/2011/06/half-belt-hack.html and later with a twist https://hydraraptor.blogspot.com/2011/12/mendel90.html. Perhaps it should just assert, rather than supporting it.

SmoothieAq:
I have made a new commit:

> We already have a slice() function with Python semantics here: 

I have deleted the new slice and uses the old one

> I don't like using .x and .y for [0] and [1] when it isn't an X or Y coordinate as it makes the code harder to read (although I may have done that myself!).

I have tried fixing that (except in the original code ;-)

> mirrored = xor(twisted[i], p.z < 0) ? 0 : 1;

yes, fixed

> I don't think we need two versions of rounded_polygon_tangents(). 

I have refac'ed to use the new one, but please review, thank you

> I don't think belt_length() should round up to a multiple of teeth. 

Fixed

> Now you have added the open option the gap stuff is redundant. Since belt_length() taking a type is breaking I don't mind breaking the gap stuff if it is straight forward to use the open option.

Well, I don't know if `open` is straight forward, you have to be the judge for that. The reason I made it is actually to support cases where you twist the belt around and clips it together to attach it to a carriage (again, see the BLV design for an example). That can not be made with the gap. Please let me know if `open` is enough straight forward that I should drop gap :-)

How to run the belt have grown minor religions the later years ;-) For the printer I have been designing to-and-from for several years I am going to start with the twist design. On my oldish Anet A6 I modded the X to have tooth on smooth idler, but have kept the Y to have tooth on toothed pulley. But hard to say what is best. On my new printer I plan to do some experiments with an adxl345 accelerometer (see https://www.klipper3d.org/Measuring_Resonances.html), and can then hopefully get a bit wiser.

nophead:
I get errors in tests/core_xy.scad.
```
ERROR: all points for rotate_extrude() must have the same X coordinate sign (range is -0.25 -> 0.38)
ERROR: all points for rotate_extrude() must have the same X coordinate sign (range is -0.25 -> 0.38)
```
In the machine I am working on I get:
```
WARNING: Problem converting rotate(a=[0, 0, nan]) parameter in file ../../OpenSCADLibs/NopSCADlib/vitamins/belt.scad, line 104
TRACE: called by 'rotate' in file belt.scad, line 104
TRACE: called by 'translate' in file belt.scad, line 104
TRACE: called by 'for' in file belt.scad, line 101
```
I think it is because I have some zero radius corners. The old code coped with that.

SmoothieAq:
> I get errors in tests/core_xy.scad.
> 
> ```
> ERROR: all points for rotate_extrude() must have the same X coordinate sign (range is -0.25 -> 0.38)
> ERROR: all points for rotate_extrude() must have the same X coordinate sign (range is -0.25 -> 0.38)
> ```
> 
> In the machine I am working on I get:
> 
> ```
> WARNING: Problem converting rotate(a=[0, 0, nan]) parameter in file ../../OpenSCADLibs/NopSCADlib/vitamins/belt.scad, line 104
> TRACE: called by 'rotate' in file belt.scad, line 104
> TRACE: called by 'translate' in file belt.scad, line 104
> TRACE: called by 'for' in file belt.scad, line 101
> ```
> 
> I think it is because I have some zero radius corners. The old code coped with that.

This have been haunting me. I get that sometimes and sometimes not, both on 2021.01 and on a newer snapshot. I will try to nail it, but would appreciate any input. Of cause it ought to work with zero radius, but I might miss a check somewhere.

nophead:
The original code tested r for 0 here: https://github.com/nophead/NopSCADlib/blob/master/utils/rounded_polygon.scad#L61 and omitted drawing circles when the radius is 0.

SmoothieAq:
> The original code tested r for 0 here: https://github.com/nophead/NopSCADlib/blob/master/utils/rounded_polygon.scad#L61 and omitted drawing circles when the radius is 0.

Yes, I tried a different fix however. Hopefully it works now...

nophead:
It doesn't get the nan in my machine but doesn't draw quite the same. It still bombs out in tests/core_xy.scad.

I think the problem is in drawing very sharp corners. The old method used offset() but the new method tries to draw a tiny arc where the back of the belt has a negative coordinate. It isn't really physically possible to bend a belt at right angles through the pitch line, it is only used as a bodge with the gap method.  I will try to convert the two problems to use the open method. If that works out then we can get rid of gap and not allow a bend radius so small it causes a negative radius at the back of the belt.

nophead:
It looks like you don't run the tests before submitting a PR, can I ask why?

As well as tests/core_xy.scad failing there are also a lot of typos detected by codespell.

```
readme.md:276: Alternativly ==> Alternatively
readme.md:285: Alternativly ==> Alternatively
readme.md:285: specifed ==> specified
readme.md:286: Alternativly ==> Alternatively
readme.md:287: automaticall ==> automatically, automatic
readme.md:309: calulates ==> calculates
readme.md:5845: recursivly ==> recursively
readme.md:5845: inital ==> initial
```

Changing core_xy to use an open belt seems to fix the problem, so I will merge this with the typo fixes and the gap stuff removed.

nophead:
Yes the open option is much better than the gap options it replaced and easier to use. Thanks for the work. @martinbudden might want to cast an eye over the changes I made to utils/core_xy.scad as he was the original author.

SmoothieAq:
Great that you merged. Sorry I wasn't aware of test.py, I should have fixed what it showed before making the PR.
I will be busy with work the next days, but I do have an idea to avoid errors when using a small radius. I'll just make a new PR, if/when I have something better...

PR #171, opened 2021-03-31
Added MGN7H and MGN9H carriages.

martinbudden:


SmoothieAq:
just now, I need a mgn9h. Please merge :-)

nophead:
I am working on it at the moment but there are issues. The test routine needs to show both carriage types, which I have implemented but carriages don't appear on the BOM, so there is no indication which carriage is used in a project.

It will be merged shortly when I have fixed that.

nophead:
Merged.

PR #172, opened 2021-03-31
Added box sections.

martinbudden:
Added box sections (aka rectangular tubing) to `tubings`. I elected to add box section to tubing, rather than create a new `box_section.scad`, but I can change that if you like.


nophead:
Yes I think it is better to have a `box_section.scad` so we have two separate tests to make them easier to fit into the big picture and the test examples can be taller. I haven't tried it but I imagine box section shorter than it is wide would look a bit odd.

martinbudden:
I've tidied this and rebased. It should be good to merge now.

PR #185, opened 2021-05-31
Use correct module name in example

elliotf:
Thank you so much for sharing your OpenSCAD library with the world! Here is a minor typo fix to the readme.md.

Also, while I'm thank you, thank you for all of your work on your Mendel90. You've been in inspiration to me for the last .. at least twelve years or so. So, thank you.


nophead:
Thanks for spotting it. readme.md is generated by tests.py, so what is needed is to correct the source code. I have done that here: 20d799a3c115d3d32f101c4419d6e9b57c3be8c7.

PR #189, opened 2021-06-15
Spelling corrections.

martinbudden:


nophead:
So much for using codespell!

Ordinate is actually correct when there is only one.

martinbudden:
Codespell seems to only run on files that are open in the editor. I expect there is a way to get it to run over all files, but I haven't looked into doing that.

I don't think "ordinate" is correct here -

From the Cambridge dictionary:
> ordinate
noun [ C ]
mathematics   specialized
the second number in a pair; a number on a Y-axis that shows the coordinate (= the place along that line) of a point

nophead:
Codespell is a command line utility that I run on the generated markdown file. Annoyingly it suggests ream should stream in my 3D printer but hasn't picked up any of these.

Coordinates have two or more values. I thought ordinate we as the singular form of coordinate but it turns out it is only the Y axis.  Apparently the z value is called the applicate but z-coordinate is more easily understood as the z component of a coordinate.

martinbudden:
I found these using the Code Spell Checker extension for Visual Studio Code, hence my confusion with Codespell.

I also found `rivit` in `button.scad`, but didn't change that because it is part of a function name and so would break compatibility. I think there may be some others, so I'll look at doing a single pull request with all spelling related changes to code.

nophead:
The latest codespell doesn't complain about rivit, or any of your other corrections, so all I can assume is Microsoft have their own dictionaries.

PR #191, opened 2021-08-09
Added ability to display camera field of view.

martinbudden:
Useful for checking camera position and angle when designing a mount for a camera.

cameraFOV


martinbudden:
That's what I've done, haven't I?. The default FOV is a rectangle specified by the camera type. I've allowed this to be overridden because you can change the lens on some cameras. I've also allowed just a single angle to be set, resulting in a conical field of view. That's not because that's how some cameras behave, it's because sometimes the datasheets only specify a single angle - but I could get rid of that option if you want.

nophead:
I think all you need is the fov angle and the image aspect ratio. It isn't one or the other. The aspect ratio is defined by the sensor and the angle is defined by the focal length of the lens. Unless the sensor is square there are two angles but you only need one to draw the rectangular cone, presumably if only one is specified there isa  convention as to which it is.

The aspect ratio can be a fixed property of the camera, only the focal length of the lens needs to be specified to draw the FOV, or one of the angles.

martinbudden:
Can we resolve this, so that this can go in with your current merge?

So it is true, that you can specify this in terms of the sensor size and lens focal length - but:

1. NopSCADlib does not have separate sensor and lens objects - if you wanted to have the raspberry pi quality camera with a telephoto lens, you'd have to specify a whole new camera object.
2. The spec sheets for the cameras tend to specify the FOV, not the sensor size or sensor aspect ration.

nophead:
I don't understand your maths. Why is it not just `2 * fov_distance * sin(fov/2)`? Why the `+(fov > 90 ? 1 : 0)`.

I don't think it makes sense to show a cone. It is always a rectangle and either you need to know two fov angles or one plus the aspect ratio of the sensor. If you only have one number you can't draw it.

nophead:
Actually it should be `2 * fov_distance * tan(fov/2)` for a flat projection with the distance measured along the axis.

martinbudden:
I've made those changes, should be OK now.

nophead:
It should be possible to write a script that generates an image at camera resolution of exactly what the camera sees, given its placement.

PR #199, opened 2021-09-18
Convert layer_height and extrusion_width to functions

rcolyer:
This will capture dynamic scope values, while still permitting a default value for these.  Preview and render performance appears to be exactly identical to within the normal jitter of the measurements.


nophead:
You would need to change all my $variables and `squeezed_wall` is no longer a constant, it would also need to be a function and anything that uses it.

How does setting `$extrusion_width` at the top of main effect the STL generation if they ignored when the file is used?

global_defs.scad
```diff
@@ -42,15 +44,15 @@ pp4_colour      = is_undef($pp4_colour)      ? "darkorange" : $pp4_colour;// pri
...
-squeezed_wall = $preview ? 2 * extrusion_width - layer_height * (1 - PI / 4)
-                         : extrusion_width - layer_height / 2 + nozzle / 2 + extrusion_width / 2;
+squeezed_wall = $preview ? 2 * extrusion_width() - layer_height() * (1 - PI / 4)
+                         : extrusion_width() - layer_height() / 2 + nozzle / 2 + extrusion_width() / 2;
 
 inf = 1e10;      // very big
 eps = 1/128;     // small fudge factor to stop CSG barfing on coincident faces.
 $fa = 6;
-$fs = extrusion_width / 2;
+$fs = extrusion_width() / 2;
```
nophead, line 53: How can I make this dynamic? I can't turn $fs into a function.

nophead:
And lines like this would all need to be functions https://github.com/nophead/NopSCADlib/blob/52e9c1d7fd97555a03ea9c029887902d28aefae5/printed/camera_housing.scad#L34

global_defs.scad
rcolyer: If you want to override $fa and $fs from the callers, then set these at the top of externally used modules with something like $fs = get_fs(); and put whatever logic you want in that function.  I specifically didn't edit these because you mentioned that your desired behavior for $fa and $fs was disrupted in some parts of the code for some usages when you switched from use to include to try to address the issue without using functions.  But this is not necessary, and you could go back to use if you like.  Setting them via a function call allows full control and makes very clear exactly where and when the caller's $fa and $fs values are being overridden, with what behavior, and where they are not overridden.  Implementing this part I will have to leave to you because the full desired behavior for this was not clarified, but that is how to accomplish it.

nophead: I have no idea what you mean. I want `$fs` to equal `extrusion_width() / 2` dynamically as it did before. I.e. it responds to $extrusion_width set at file scope or on the stack.

nophead:
Won't merge because it requires all projects to be rewritten in a more ugly and bug prone form.

PR #201, opened 2021-09-28
Removed hotend global enums.

martinbudden:
Got rid of the global enums `jhead` and `e3d`. A global hotend enum is not very desirable.

### DIFF
--- a/global_defs.scad
+++ b/global_defs.scad
@@ -73,11 +73,6 @@ hs_grub   = 4;     // pulley set screw
 hs_cs_cap = 5;
 hs_dome   = 6;
 //
-// Hot end descriptions
-//
-jhead   = 1;
-e3d     = 2;
-//
 // Face enumeration
 //
 f_bottom = 0;
--- a/tests/hot_ends.scad
+++ b/tests/hot_ends.scad
@@ -19,18 +19,13 @@
 include <../utils/core/core.scad>
 use <../utils/layout.scad>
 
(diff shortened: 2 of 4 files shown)


nophead:
You seem to have got rid of the pseudo polymorphism. That would break my current 3D printer which has multiple extruders that vary by hotend type and I will also add two more types soon.

flexstruder_E3Dv6_175_assembled_tn

flexstruder_JHeadMk5_175_assembled_tn

nophead:
It really needs module literals, like function literals, and then we could get rid of the type enum. But while there isn't I use it to do the case by case switching.

martinbudden:
> But while there isn't I use it to do the case by case switching.

When I need to do this sort of thing I just a string literal rather than an enum. So `"e3d"` and `"jhead"` rather than `e3d` and `jhead`.

`if (hot_end_style(type) == "jhead")` is just as readable as `if (hot_end_style(type) == jhead)`

nophead:
Yes it is just as readable but you don't get an error if it you spell it wrong.

martinbudden:
> Yes it is just as readable but you don't get an error if it you spell it wrong.

You do if you spell check your code.

nophead:
True, but I don't. 

I have no objection to making them strings though. Although it is breaking I doubt anybody else is using it.

martinbudden:
Now updated as per discussion.

PR #206, opened 2021-11-16
Removed need for supports from drag chain.

martinbudden:
link

1. Made pin conical, so it doesn't require supports
2. Used conical horihole in outer cheek for pin
3. Removed cutout in inner cheek
4. Used shape of cheek circumference to limit rotation of links
5. Make clearance a property of the drag chain so it can be altered for printing
chain assemblies
6. Altered teardrops to allow extra width of top.

I've successfully printed a chain of 2 links as a single part by setting `clearance = 0.2`, but this my vary according to your printer.

The use of the cheek profiles to limit the rotation of the links can be seen below:
DragChain

### DIFF
--- a/printed/drag_chain.scad
+++ b/printed/drag_chain.scad
@@ -152,7 +152,7 @@ module drag_chain_link(type, start = false, end = false, check_kids = true) { //
                                             rotate(180)
                                                 teardrop_2d(r = r, extra_x = 2 * clearance);
                                         translate([0, r])
-                                            square([r, r / 2]);
+                                            square([r, r / 2.5]);
                                     }
 
                                     translate([outer_end_x - eps, 0])
--- a/printed/drag_chain.scad
+++ b/printed/drag_chain.scad
@@ -152,7 +152,7 @@ module drag_chain_link(type, start = false, end = false, check_kids = true) { //
                                             rotate(180)
                                                 teardrop_2d(r = r, extra_x = 2 * clearance);
                                         translate([0, r])
-                                            square([r, r / 2.5]);
+                                            square([r, r / 2]);
                                     }
 
                                     translate([outer_end_x - eps, 0])
@@ -160,10 +160,10 @@ module drag_chain_link(type, start = false, end = false, check_kids = true) { //
                                 }
                             // Conical horihole for pin
                             if(!start)
-                                translate([socket_x, r, -side * (wall + clearance) / 2])
+                                translate([socket_x, r, -side * wall / 2])
                                     hull() {
                                         horihole(r = pin_r, z = r, h = eps);
-                                        translate_z(side * (pin_h + clearance))
+                                        translate_z(side * pin_h)
                                             horihole(r = pin_r - pin_h, z = r, h = eps);
                                     }
                         }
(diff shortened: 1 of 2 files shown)


nophead:
Let me know when you have finished and I will try printing my tiny chains. I chose an actual pin because I think it makes a smaller chain. I haven't got much room.

martinbudden:
Well, I've finished this now and deem it a mixed success. The conical pin works well, but using the cheeks to limit the link rotation works less well. There are noticeable stops at 0 degrees and 45 degrees, but you can force the chain beyond those angles. The stops are probably good enough, but I'd appreciate your perspective.

On the subject of the pin, it would easily be possible to make the chain configurable to use a cylindrical or conical pin, so that the cylindrical pin could be used for small chains if required.

On the subject of the link rotation, one possibility would be to print the links the right way up, so that the cutout was on top, and so did not require supports. Is there any reason why you did not do it this way originally?

nophead:
I think all the changes need to be reversible because I need the original design for my 3D printer. My side walls are only 0.9mm thick, which is why I chose pin over dimple and why I used a cam to limit movement and not the cheeks. My design was actually based on a commercial one I have, but shrunk down greatly.  My chains are unsupported and of arbitrary length because the printer is fully parametric. So I need them to be rigid enough not to sag much.  I don't think the cheeks will do that and the commercial design must have come to the same conclusion. 

For large machines your design is probably better as you can have thicker cheeks. Mine are tiny because they only house a six way ribbon cable folded in half.

x_axis_assembled

Not sure what you mean about printing them upside down. I can't print round curves upside down as they approach infinite overhang.

martinbudden:
What I meant by the right way up was printing the links like this:

DragChain2

that is with the roof and the cutout on the top. (Note that I have inverted the teardrops, so the there is no overhang problem with the round curves).

Printing them this way means the only support required is for the pin, and for larger links a conical pin with no supports can be used.

nophead:
Before your changes the links look like this.

x_drag_chain_link

They are rounded on the top so that when they bend they don't have pointy bits sticking out at the corners like yours do.

martinbudden:
OK, I understand why you chose that orientation.

However, I don't think the "pointy bits" are problematic - they cause the chain to lift in the direction it was lifting anyhow, see below
DragChainCorner

nophead:
Yes it isn't a big problem. Ten years ago I made square edged ones and used the cheeks to limit movement and they worked fine on a big heavy machine, but they are bit clunky.

IMG_20211118_174932407_HDR

However, for this much more delicate machine I spent the time to optimise the design and printed a few and tested them. I don't really want to go backwards. I don't normally use cable chains on 3D printers because they are overkill. I prefer to use polypropylene strips and ribbon cable where I can. For this machine I just couldn't find the space, so I had to resort to two very light weight cable chains on X & Y. 

image

Ironically the Z-axis, which is the strongest and slowest moving can use a strip.

image

So I think it needs to be able to create the original shape. It will pretty complicated if there is the option to print it the other way up.

martinbudden:
So I think the idea of using the cheeks to restrain the link rotation is a dead end.

I've submitted a V2 of this proposal, which uses the original constraint mechanism. In V2 the original chain is printed by default, but if `supports` is set to false, the a conical pin and hole is used, the teardrops are inverted, and the chain is printed with the cutouts on the top (so they don't require supports).

nophead:
I am assuming this is superseded and can be closed?

martinbudden:
Yes, closing.

PR #209, opened 2021-11-27
Improvements to core_xy

martinbudden:
1. Better default parameters
2. Changed plain_idler_offset to work in x and y, to facilitate larger idler pulleys

This example uses 25 tooth idler pulleys - they have larger and wider bearings than 20 and 16 tooth pulleys and are hence more durable.
CoreXY

### DIFF
--- a/tests/core_xy.scad
+++ b/tests/core_xy.scad
@@ -30,22 +30,25 @@ module coreXY_belts_test() {
     plain_idler = coreXY_plain_idler(coreXY_type);
     toothed_idler = coreXY_toothed_idler(coreXY_type);
 
+    pos = [100, 50];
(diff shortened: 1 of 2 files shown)


nophead:
Is there a bit of the green belt missing?

martinbudden:
If you zoom in you can (just about) see that the green belt is hidden behind the blue belt.

Here's the same setup from a slightly different viewpoint
CoreXY2

nophead:
Looks like the test needs updating to put the screws in the right place.

image

martinbudden:
OK fixed now. Turns out this revealed a minor bug in the corexy y_carriage plain pulley position, which I've also fixed.

PR #210, opened 2021-12-10
Enhanced PSU to allow terminals at both ends.

martinbudden:
For example
psu
PSUscad

### DIFF
--- a/vitamins/psu.scad
+++ b/vitamins/psu.scad
@@ -182,8 +182,13 @@ module psu(type) { //! Draw a power supply
                                         polygon([for(p = cutout) p]);
 
                                 for(h = psu_face_holes(f))
-                                    translate(h)
-                                        drill(screw_pilot_hole(psu_screw(type)), 0);
+                                    translate([h.x, h.y])
+                                        hull() {
+                                            drill(screw_pilot_hole(psu_screw(type)), 0);
+                                            if (is_list(h[2]))
+                                                translate([h[2].x, h[2].y])
+                                                    drill(screw_pilot_hole(psu_screw(type)), 0);
+                                        }
 
                                 g = psu_face_grill(f);
                                 if(g) {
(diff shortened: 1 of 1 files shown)


nophead:
This change makes the PCB too long at the front and the back. It used to look like this:

image

It now looks like this:

image

I also noticed you removed the square cutout in the corner of the S-250-48, which seems to be correct but the terminals look to be further to the right and the left wall seems like it should be longer.

image

I am away from my hardware again this winter, so I don't have anything real to look at, even though I have one of those at home in the UK.

nophead:
I also noticed the shroud is at the wrong side for the  S-250-48.

image

Not related to this change but when this PSU was added,

nophead:
Actually it is the S_300_12 that looks wrong but it seems they are available in many styles, some with the mains on the left and some on the right. Some with the corner cutout for screwdriver access to a slot below and some with the connector at the edge.

martinbudden:
Yes, I think the S_300_12 is wrong - it does seem most have the shroud on the other side. I can correct that in another PR.

For this PR I'll update it so that there is no change to the existing PSUs.

martinbudden:
Further thoughts - I've been thinking about how to proceed with this.

Rather than add support for terminals at both ends of the PSU, I think a better approach is to do the following:

1. Add support for terminals on PCBs. This is actually useful in its own right, I want to do this for a separate reason anyway.
2. Add the ability to add a PCB to a PSU, the terminals would then be on this PCB. (This also allows the PCB to be more fully populated, so that the innards of the PSU could be shown, in so desired).

What do you think? If you are in agreement then I'll close this PR and (in due course) open new ones for (1) and (2) above.

nophead:
Yes that seems like a better idea than the original PSU terminals. All PSUs have a PCB, so why not model it as a PCB with at least terminals on it.

martinbudden:
Withdrawn, as discussed.

PR #211, opened 2021-12-11
Allow configuration of teardrop angle.

martinbudden:
In most cases a teardrop angle of 45 degrees is fine, but there are cases when this does not give sufficient overlap between layers and a slightly steeper teardrop angle, say 50 or 55 degrees is desirable. This change allows this. 

The teardrop angle is defaulted to its current value of 45 degrees.

In particular I print mostly functional parts and for these (for both strength and speed of printing) I tend to use a 0.6mm nozzle with 0.5mm layer height and 1.0mm extrusion width. For these large layers, a teardrop angle of 50 or 55 degrees gives better results.

### DIFF
--- a/global_defs.scad
+++ b/global_defs.scad
@@ -35,6 +35,7 @@ cnc_bit_r       = is_undef($cnc_bit_r)       ? 1.2    : $cnc_bit_r;       // min
 show_rays       = is_undef($show_rays)       ? false  : $show_rays;       // show camera sight lines and light direction
 show_threads    = is_undef($show_threads)    ? false  : $show_threads;    // show screw threads
 show_plugs      = is_undef($show_plugs)      ? false  : $show_plugs;      // plugs on headers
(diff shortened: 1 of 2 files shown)


nophead:
I normally merge using the command lines provided by the PR. It worked yesterday but I don't see any today on the remaining PRs. Even the PSU one I am part way through merging.

martinbudden:
I'm not sure what you mean - is there some kind of problem with github?

nophead:
Yes either GitHub has changed overnight or I have gone blind

nophead:
With a diagram I worked out x is simply tan(teardrop_angle /2) and Wolfram says that is equivalent:

image

So the top of the teardrop is 2 R tan(a/2), where have I seen that before today, LOL!

I will simplify it when I merge. It's a pain now Github doesn't give the commands anymore.  I have to type them all manually.

PR #212, opened 2022-01-07
Fix grid origin offsets

ggabbiani:
Hi all,

imho the origin offsets of the perfboard are wrong. The fixed values should be the result of the following calculations:

    (<size> - (<number of holes> - 1) * <step>) / 2

So for a perfboard 80x20mm with 28x6 grid holes the results are

    (80 - 27 * 2.54) / 2 = 5.71
    (20 - 5 * 2.54) / 2 = 3.65

I realized this using your pcbs data in a project of mine ([OFL](https://github.com/ggabbiani/OFL)) that is using a unified algorithm for grids (this pcb integration is still a work in progress).

Regards
Giampiero


nophead:
Hello,
   The reason it isn't a formula is because the SRPB boards I modelled first are not symmetrical, the grid is actually offset. 

I think the fibreglass boards are symmetrical and I calculated the PERF60x40 and then copied it to the others thinking they look the same. But because the sizes are in mm and the pitch in inches they are all slightly different. I have made several projects and not noticed any problems but the changes are pretty small, except got the SRPB one. I think I measured it and it is correct for the boards I have. I can't check because I am at my winter residence and haven't got any here.

If you revert the PERF74x51 change I can merge it.

ggabbiani:
Hi,
just reverted the PERF74x51 as it was.

Regards
Giampiero

nophead:
Thanks.

PR #214, opened 2022-01-23
Added ability to add PCB to PSU

martinbudden:
This requires support for terminals on PCBs to be merged first.


nophead:
I think the part number in the description should not have the NG prefix as a Google search with that does not find them but without it does. https://www.aliexpress.com/item/1005003739241723.html

PR #235, opened 2022-07-03
Add SOIC SMD Package

MS3FGX:
This is modification of the existing SMD transistor functions/code to render Small Outline Integrated Circuit (SOIC) packages. I've also included SOIC8, SOIC16, SOIC18 components, with the dimensions coming from combination of my own measurements and datasheets.

soic_demo

I realize this duplicates a lot of the code from smd_sot, and a more appropriate long-term approach might be a more generic function that could handle multiple package types. Something like `smd_dev(package, type, value)` But it seemed like this most direct way to add new part without breaking backwards compatibility.


nophead:
Great. It has been on my to do list since I added rounded_path to make the gull wings easier.

The gull wings on the larger packages don't look quite right though.  The starts and ends seem to small.

image

MS3FGX:
Yeah, looks like I had body width too large for the SOIC18, which was making the leads look smaller. I've also lowered the lead height a bit for SOIC18 which makes the angle less aggressive, and changed foot depth to .5 for all the chips for better visual balance. Overall I think this brings the look a little more in line with the SOT components. 

soic2

PR #239, opened 2022-09-27
Leadnuts: add nuthousing

jeroenrnl:
After the previous simple one, I decided to take up something a little more challenging and I've added an aluminium housing for leadnuts.
Again, please feel free to let me know if you'd like anything done differently.

leadnuts


nophead:
Thanks, Never knew such a thing existed, I 3D print mine. Could be an option if I need a stiffer one but I would have to drill and tap more holes.

PR #245, opened 2022-10-20
update nuts.scad table comment

drf5n:
Change comment about nut sizes from radius to diameter.


nophead:
Why make_all.py? I just type make_all so make_all is the command, make_all.py is the script that implements it.

drf5n:
I see-Windows omits the extension that one would need on Mac and Linux.

drf5n:
I backed out the make all commit from this patch.  Now it's just a comment update.

nophead:
Thanks.

PR #246, opened 2022-12-05
Include terminal.scad in lib.scad to allow use of terminal_block().

majekw:
Add include of vitamins/termian.scad to lib.scad to allow using terminal_block() with only include of NopSCADlib/lib.scad.


nophead:
Terminal blocks are normally only used on PCBs or PSUs. Are you using it for something else?

majekw:
I just draw PCBs 'by hand', not using pcb modules.
I use also JST, pin_header, green_terminals, TO247, FPC and all are available also outside of pcb, so I was thinking that for consistency terminal_block should be also available globally.

PR #248, opened 2023-01-23
Add Raspberry Pi 3 A+ to pcbs

elliotf:
Thank you for this excellent library! I did not see a "contributing" document, so please let me know if there is anything I missed.


nophead:
Thanks. The only bit you missed was adding it to the list of pcbs so it gets included in the test and the images but that is no problem for me to add.

PR #250, opened 2023-01-28
WiP of an Orange Pi Zero 2 PCB

elliotf:
Hello again, nice people/person!

Since Raspberry Pis are so hard to come by these days, I've picked up an Orange Pi Zero 2, and I've added it as a PCB.

There are a few issues with it, though, mostly centered around the vertical USB A socket:
* The dimensions of the socket are off from what is in NopSCADlib, and (unlike the microsd socket) there is no way to alter the dimensions.
* Related, I started adding an ability to override the dimensions via a proof-of-concept "usb_vAx1" vertical USB A socket that I do not much care for, so thought it would be good to get some feedback before I proceed further.
* The socket does not have any 90 degree flanges. I can turn off some of the flanges, but not all.

Thank you in advance for any guidance you might have, and thank you again for this library!

### DIFF
--- a/vitamins/pcb.scad
+++ b/vitamins/pcb.scad
@@ -115,6 +115,14 @@ module usb_A_tongue() {
                 }
 }
 
+module usb_vAx1(cutout = false) { //! Draw USB type A single socket
+    translate([0,6.5/2,13.25/2]) {
+      rotate([90,0,0]) {
+        usb_A(h = 6.5, v_flange_l = 0, bar = 0, cutout = cutout, 14);
+      }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


gounthar:
I can't help, but thank you so much for  adding this board.

nophead:
I have made the flanges removable and fixed the tongue position, just the overall dimensions to sort out. As a sanity check, what is the wall thickness of the connector?

elliotf:
The thickness of the metal is ~0.3mm if that's what you're asking.

nophead:
Yes that makes sense. It is thinner than the original.

Issue #240, opened 2022-09-27
SBR16 rails


jeroenrnl:
I am working on putting a design of a CNC machine in OpenSCAD and I'm trying to supply this project with anything that I'm missing.

The next part, I'd like to model is SBR16 rails (and possibly SBR12 and SBR20 if I can find detailed enough pictures / measurements online as I do not have any samples myself).

![](https://ae01.alicdn.com/kf/S4ea97f6997d9406db37c343fab45a5bbl.jpg)

I'm just doubting where I will put the parts as they fit in a few categories...
### rails ###
This is essentially a steel rod on a mounting rail. I could put it under *rods*, function wise it would better fit in *rails*, however, both the implementation for the rail and the carriage would not fit in the current functions so ```rail(SBR20, 900)``` wouldn't work without significant overhauling of the current implementation. A third option would be creating a new *SBR* category.

### carriage ###
The carriage is really a bearing block, but with a cutout for the rail mount:
![](https://ae01.alicdn.com/kf/HTB1OXIbc.WF3KVjSZPhq6xclXXa7.jpg)
So it could be an addition to *bearing block*, but it could also be added to *rail* or *SBR* if that's the choice above.

I think adding it to *rail* makes the most sense, from an end-user point of view, but it will require some refactoring of the current code. I'd like to hear your take on it before I start ~destructtion~ making changes.

nophead:
I think the bearing blocks are similar enough to the existing bearing blocks that can be added by adding a split option. Then I think we need a new class for the rail. I think it is probably too different to shoehorn into the existing rails. The only downside of having a new class is that you can't easily make a machine that can switch from one rail type to another but that is probably an early design decision.

nophead:
Added via #258, thanks.

PR #258, opened 2023-06-27
Added SBR Rails

jeroenrnl:
Issue #240
This change adds
- SBR rails. For now only SBR16
- The bearing block that goes with it (SBR16UU). I made a split in `bearing_blocks` to support a new kind of bearing block. I tried to reuse some code, but due to the fact that the letters used to indicate measurements are vastly different between SBR and SCS - well and the fact that the blocks are quite different as well - this was only possible in a very limited way.
- An 'open' linear bearing (LM16UUOP) that is used by the bearing block, I made some changes in the `linear_bearing[s].scad` to support this.
- A 28mm circlip that has a wider opening than then normal circlip.
(circlip_28iw)

preview

It should be fairly easy to add the other types of rails, bearing blocks and linear bearings, but I first wanted to hear whether my current approach is the right one.
I used [this PDF](https://cnc4you.co.uk/resources/SBRxxUU.pdf), as well as a set of SBR16 rails I had on hand and several other online drawings I could find.

The 28mm 'wide' open circlip is something I could not find any documentation on, so the measurements are mostly guessed, based on the one that is in the SBR16UU I have. The name '*circlip_28iw*' is something I made up.


vitamins/bearing_block.scad
```diff
@@ -169,3 +170,147 @@ module scs_bearing_block_assembly(type, part_thickness, screw_type, nut_type) {
...
+
+    translate([0,-part_thickness,0])
+        sbr_bearing_block_hole_positions(type)
+            rotate([0,180,0])
+                screw(screw_type, screw_length);
+}
+
+/* Generic bearing block functions */
+
+module bearing_block_hole_positions(type, y) { //! Place children at hole positions
```
nophead, line 303: This doesn't seem to be generic. All the scs blocks now have their screws in the sbr positions, which is wrong.

nophead (commented):
Looks good apart from the SCS screw positions.

vitamins/bearing_block.scad
jeroenrnl: Weird I didn't catch that, as I did check the results on the SCS blocks... Looking into it.

jeroenrnl: > This doesn't seem to be generic. All the scs blocks now have their screws in the sbr positions, which is wrong.

It should be fixed now. I removed the generic function as after the change the functions had very little in common now.

jeroenrnl:
I also rebased the branch on master so it has the changes you made in the past few days incorporated.

nophead:
I took the liberty of deriving the open angle from the open width. I can't see how they can not be exactly related. On the data sheets the numbers vary a bit but that seems to just be rounding as only one can be an integer.

PR #261, opened 2023-11-03
Add files via upload

emteeoh:
The most basic form of a Mini-ITX PC motherboard. The grey cube is the IO cut-out . The exact placement of the PCIe slot is missing but it would be inline with the 2 mountpoints on the left of the board.


nophead:
Hi, l haven't got a PC to look at this ATM. I might be in a position to look at it in a week.

emteeoh:
I apologise for the crappy PR name.  This creates a very basic shape for a mini-ITX motherboard with a block for the IO shield cutout.

nophead:
The hole positions don't look correct relative to the edge of the board.

To be included in the library it needs to modelled as a PCB.

PR #262, opened 2023-11-12
Add ESP-01M WIFI Module to PCBs

xrxca:
I use these in a few projects, so added a definition for it to the pcbs.scad file.

Dimensions were taken from the board specs so should be accurate.


nophead:
I merged this, thanks. The only problem was your commit comment just said "Update pcbs.scad" and the changelog is built automatically from the commit comments and the semantic version tags. I tried to add a commit that mentioned ESP-01 but now it looks like I added it for some reason it says you then updated pcbs.scad three times. It's a bit of a mess. Please use a descriptive commit comment if you do another PR.

PR #267, opened 2024-02-01
Update global.scad

Minimoed:
Replaced the following functions:
foot(x) = x * 25.4 * 12; with foot(x) = inch(x) * 12; yard(x) = x * 25.4 * 12 * 3; with yard(x) = foot(x) * 3;
m(x)    = x * 1000.0;  with m(x) = cm(x) * 100;


nophead:
Why, it will juts make it slower to execute as yard() will be three function calls?

PR #270, opened 2024-02-04
Add extra MDF thicknesses

jeroenrnl:
This adds two fairly common MDF thicknesses of 3 and 22mm

PS I don't know why the scripts always seem to change "PCB" in "pcb" or "Pcb"


nophead:
The PCB case problem is due to PCB_utils.scad having the wrong case on github but the correct case on my Windows machines.  I Have just fixed it on Github using the web interface so it should go away.

PR #272, opened 2024-02-13
Add threaded inserts

jeroenrnl:
I've added some threaded inserts to be used in wood.
I took the measurements from a drawing I found online, but the thread pitch for the outer threads were just made up, there are so many different types that it doesn't make much sense to replicate a specific one.

image


jeroenrnl:
Hold this one for a moment: I found a document with much more precise measurements for the DIN standard for these inserts. Including thread pitch, etc. I will adept to that and push an update later.

jeroenrnl:
https://www.fasteners.eu/standards/DIN/7965/

nophead:
Would be great if you could add M4 as I actually use those, so far not on anything I needed to model.

nophead:
Do we need such large parts in the library? I am struggling to find anybody that sells them. What sizes are you actually using?

nophead:
Also the biggest screw in the library is M8 at the moment. I think the biggest screws I have ever used anywhere are only M10 as the tensile strength of an M10 screw is enormous, something like 2300kg. why would you use such a big screw in wood?

jeroenrnl:
To be honest, I haven't really given the uses much thought, I just copied the data that I had. I mainly use M3/M4/M5/M6 in my projects, but haven't used these inserts much, but I'm working on a project now that requires M6, so I modelled it.

What I can think of for the bigger sizes, is using them with threaded rods to make some kind of screw-adjustable jig. My local tools and materials supply stocks these up to M20. For some of these jigs M14 and M16 have the advantage of having a nice pitch of 2mm so you can achieve a nice turns-to-advancement ratio.

Anyway, it's up to you to decide, as long as you keep the most used ones, it's fine by me. :-)

PR #273, opened 2024-02-18
Added T-Tracks and accessories

jeroenrnl:
I have added T-Tracks with T-bolts and inserts.

Unfortunately, the sizing of these things seems far from standard. I have added what I could find as "common" sizes.

image

Unfortunatelty, I couldn't find the sized for the inserts for the 36 mm track. I could quite simply create this by looking at the size differences between the 30mm and 36mm tracks, but there's no guarantee that is accurate.

I also have a few of these modelled as 3D prints, but they need a little cleanup before I can add them. I will do that some other time.


jeroenrnl:
Oh, and the screws used for mounting them are now bolts instead of wood screws.
I would really like to add some metric size woods screws too in the near future.

nophead:
Is there any difference between metric wood screws and imperial? I think they are just labelled in metric.

PR #274, opened 2024-03-14
Update MT3608 pcb module

vshcryabets:
Hello, i think it can be helpfull.

I've updated MT3608 PCB
Before:
Screenshot from 2024-03-14 20-52-11

After update:
Screenshot from 2024-03-14 20-51-02

Also I've added CDRH104 inductor (it's used on MT3608)
Screenshot from 2024-03-14 20-51-41


nophead:
Looks nice, thanks.

PR #281, opened 2024-08-14
Dome screw calculations fix and additional vitamins 

allTexas:
Howdy! 

First time pull request, so please forgive me if I missed part of how this is normally done or if some of these changes ought to be done differently.   I made a few updates to the incredible NopSCADlib repo so it fits even more of my needs.  I also improved (in my humble opinion) the math for calculating the curves on dome screws -- it now treats the flat spot on top as a  sphere cap, and uses the equations for such to calculate the curvature -- with mathematically equivalent ratios of square roots instead of trig calls for processing speed.  This change allows larger and smaller sizes than were possible with the old math, and also eliminates the need for that "lift" value and the special handling of M2 sized screws.  I also added M6 and M8 screws (in part to show off and in part because I'm using M6 screws in a current project) that were not in the screws.scad before. 

Main additions are: 

      weld nuts,
      
      F625 ball bearings 
      
      Rod end bearings
      
      a handful of Idler and pulley sizes that I've used in 3d printers and wanted to have modeled. 
      
      Aforementioned M6 and M8 dome screws 
      
 Again, if there's that needs changed before merging or you want some commits and not others that's fine.  Also a huge thank you to the team that's made this amazing repository.

--
Tex


nophead:
I wrote the dome code a long time ago but I think I did it to match the screws I have. I haven't managed to find an online definition for the shape and I suspect it varies a bit by manufacturer.

One place I buy screws from have STL models but they don't match the screws they sell. They have a sharp edge at the base like your version but my screws don't. 

WIN_20240816_00_00_54_Pro

I will see if I can get my head around the maths tomorrow. Do you screws actually look like your model?

allTexas:
Ahh, that little radius at the bottom of the button head was the origin of the 'lift' value I'm guessing?  I do believe that fine of detail varies quite a bit between manufacturers, especially since it's not in their drawings or models.

I usually buy bolts from McMaster or boltdepot, the M6 I'm currently using on one project are these:
https://www.mcmaster.com/92095A242  both McMaster and boltdepot both have sharp corners shown in their drawings, but in real life both have some amount of radius on the thread side of the head.  I'm guessing the radius is either a manufacturing artifact, or perhaps the sharp edge is smoothed intentionally to avoid cutting fingers during handling.  

Perhaps we add back in a lift component, (to represent that part that's non-dome shaped at the bottom), but make it a percentage of the head_height so that it scales nicely to the larger sizes? Ie, having the lift always be 0.38mm would be hardly visible on a huge M16 screw, where the bottom side rounding/radius appears to still occupy the bottom ~10% of the head height. https://www.mcmaster.com/91306A755

nophead:
It isn't just a radius. It is a mostly flat section with tiny radii at its margins. Sometimes it looks like it has a negative slant as well. Yes is should probably be a percentage of the head height as you say. Also it shouldn't go wrong on the larger sizes so I need to fix that bug. I will try to work out what my maths was trying to do.

nophead:
I scanned some screws and found the radius of the flat section on the top seems to 1.3 times the across flat dimension of the socket.

Scan_20240816

Making the edge height equal to the head radius / 7.5 gives the 0.38mm height for M3 screws which is probably what I measured.

That gives two points on the arc of the dome and where a perpendicular bisector of the line between the two points hits the Y axis is the centre of the dome.

```
        if(head_type == hs_dome) {
            edge_height = head_rad / 7.5;
            p0 = [head_rad, edge_height];                   // Lowest point on the arc
            p1 = [1.3 * socket_rad / cos(30), head_height]; // Highest point on the arc
            p = (p0 + p1) / 2;                              // Start of bisector
            gradient = (p0.x - p1.x) / (p1.y - p0.y);       // Gradient of perpendicular bisector = -1 / gradient of the line between p10 and p1
            c = p.y - gradient * p.x; 						// Y ordinate of the centre of the dome
            r = norm(p1 - [0, c]);							// Dome radius is distance from centre
            color(colour) {
                rotate_extrude() {
                    difference() {
                        intersection() {
                            translate([0, c])
                                circle(r);

                            square([head_rad, head_height]);
                        }
                        translate([0, head_height - socket_depth])
                            square([socket_rad, 10]);
                    }
                }
                linear_extrude(head_height)
                    difference() {
                        circle(socket_rad + eps);
                        circle(socket_rad, $fn = 6);
                    }
            }
```

That gives screws that look about right to me.

image

What do you think? I haven't got any of the larger sizes you added.

allTexas:
Yeah, that looks like a good way to do it.  On further inspection, I noticed that the 'flat' spot is angled in slightly towards the bottom of the head on both your bolts and mine.  
image

What if we do something like this:

      ```
     if(head_type == hs_dome) {
            edge_height = head_rad / 7.5;
            head_chamfer_x=edge_height/2;
            head_fillet_radius= 0.5;
            p0 = [head_rad, edge_height];                   // Lowest point on the arc
            p1 = [1.3 * socket_rad / cos(30), head_height]; // Highest point on the arc
            p = (p0 + p1) / 2;                              // Start of bisector
            gradient = (p0.x - p1.x) / (p1.y - p0.y);       // Gradient of perpendicular bisector = -1 / gradient of the line between p10 and p1
            c = p.y - gradient * p.x;                       // Y ordinate of the centre of the dome
            r = norm(p1 - [0, c]);                          // Dome radius is distance from centre
            color(colour) {
                rotate_extrude() {
                    difference() {
                        intersection() {
                            translate([0, c])
                                circle(r);

                            // square([head_rad, head_height]);
                            offset(head_fillet_radius) offset(-head_fillet_radius)
                            polygon(points = [
                                [0,0],
                                [head_rad-head_chamfer_x,0],
                                [head_rad, edge_height],
                                [head_rad,head_height],
                                [0,head_height],
                                ]);
                        }
                        translate([0, head_height - socket_depth])
                            square([socket_rad, 10]);
                    }
                }
                linear_extrude(head_height)
                    difference() {
                        circle(socket_rad + eps);
                        circle(socket_rad, $fn = 6);
                    }
            }
            color(colour * 0.9)
                translate_z(head_height - socket_depth)
                    cylinder(h=2 * eps, r=socket_rad, $fn = 6);
            shaft();
        }
```

allTexas:
It's basically your updated logic from this morning for the dome, but with the intersection mask being a polygon instead of a square.  That way two points of the polygon can make that flat spot angled in towards the bottom of the head.

And I'd be fine leaving out the offset line that performs the radius, that might be too much computation for bolts now that I think about it.   

image

nophead:
Looks good but I think the chamfer angle is only about 15 degrees from vertical. I printed this picture I found on the web and measured it with a protractor.

s-l1200

There is a round module in the library that does the double offset. Don't worry about the speed. All 2D operations are very fast.

nophead:
Regarding the stepper motor. Does it have an 80mm plain shaft or is it a lead screw? If the latter you can specify the lead pitch and the number of starts to be drawn as a lead screw.

Any functions or modules you have added that are called by the use need comments starting with `//!`  to make the documentation in the readme.

The rod_ends and the new nuts need test code to make the pictures and documentation.

allTexas:
The steppers have an 80mm plain shaft, they look like this: https://phaserfpv.com.au/products/ldo-42sth48-2004mah

I  just added test code for the weld nuts and rod ends -- tried to copy the format from some of the existing vitamins.   There really is an impressive amount of automation in this library, I didn't realize the documentation was all auto generated based on comments.  Very neat feature!  I'm still learning all the magic here, so I appreciate you helping point out what I didn't know I missed

I also modified the dome screws code to now have the flat spot on the bottom of the head calculated by angle, and the angle set to 15 degrees from vertical.

nophead:
Thanks. 

I fixed the colours of the rod ends. The bearing shield colour was being applied to all of the round bit. I assumed it should be silver except for the bearing shields.

PR #283, opened 2024-09-25
Corrected HGH20CA carriage_pitch_x value.

jijamik:
carriage_pitch_x should be 36 instead of 35 for HGH20CA carriage
image


nophead:
That carriage was contributed by Niclas Hedhman, not sure how he didn't notice if he was using it.

PR #286, opened 2024-10-05
added HT waterpipes

nerdyjan:
added HT waterpipes from common hardwarestores, great for building cases for IoT things ;)


nophead:
Looks good but it needs a test file that invokes all the modules. That  makes the picture and documentation in the readme when tests.py is run.

nerdyjan:
i will add that

nerdyjan:
done

nerdyjan:
please wait with merging as i still working on the HT parts, will add ALL straight parts and caps just to be more complete, maybe i find the need for my current project to add all the archs and other fittings.
I don't know if these pipes are common outside germany but they are a wonderfull, cheap posibillity to create all kind of "things" ;)

nophead:
I can buy Ostendorf HTsafe HT Pipe from Amazon in the UK. It seems to be a successor with better fire rating. Is it the same dimensions, just a different material?

The test file should be named HT_pipes.scad to get the title capitalised correctly. If you not on Windows I think you can easily change it. I am not sure how to do it on Windows because git is case insensitive on Windows. I changed my local copy here but git doesn't see a change.

nerdyjan:
ok, now i'm done with the straight pipes .. maybe add more later, but no time right now. please feel free to merge.

nerdyjan:
i'm NEVER ever on windows ;) ok, will change the name

nerdyjan:
> I can buy Ostendorf HTsafe HT Pipe from Amazon in the UK. It seems to be a successor with better fire rating. Is it the same dimensions, just a different material?
i think it's the same PVC type ... i'm btw the founder of the former German RepRap GmbH ;)

> 
> The test file should be named HT_pipes.scad to get the title capitalised correctly. If you not on Windows I think you can easily change it. I am not sure how to do it on Windows because git is case insensitive on Windows. I changed my local copy here but git doesn't see a change.

Don't ask me about code from redmond!

nophead:
I think it is Polypropylene now. Perhaps that is the difference.

I bought quite a lot of ABS filament from RepRap GMBH in 2013. In fact all the  filament for the Mendel90 kits we sold came from Germany. Most of it from GLI Concept GmbH.

nerdyjan:
Try gluing them with "Tangit", if it works, it's a PVC type plastic, PP and PE are some of the most challanging materials to glue beside PTFE.
I developed a lot of filaments with German RepRap in coop with Perlon ... but this has all changed ... if you are interested, we can get in touch privatly.
The early ABS of german Reprap wasn't our development, it was made by Orbi-Tech in Germany.

However, i love you lib since the bginning but now i have more commercial CAD projects where it fit's perfectly. I have a lot of potential ideas to enhance the documentation "problem" and are currently coding a new kind of "static" html help with parts catalog and lib documentation generated from tests.py as by growing this lib it's getting more and more annoying to scroll the bis readme.html ... i will contribute if finished.

For this PR that's it for now but can be extended in the future with all it's archs and other fittings.
Thank you

nophead:
Yes I also bought filament from Orbi-Tech as well. The reason I used ABS from Germany is because the "natural" colour was white and didn't stink when extruded whereas all other ABS stinks and the natural colour is cream.

Note sure what you mean about scrolling the readme? It has a table of contents to jump straight to the section you want and Top links to get back to the TOC.

I took the liberty of changing a lot of your code. Rather than differencing 3D cylinders to make the pipes I use the tube() utility that is an extruded ring, which is faster and avoids the Z fighting at the bottom of the pipe. It also avoids clashes between negative objects in differences when vitamins are assembled together. That can cause random Z-fighting in assembles.

The make the HTpipeFitting I rotate_extrude a 2D profile.  While doing this I noticed your version didn't have the correct internal shape because it was solid and just got the pipe_id cylinder subtracted from it. I also made the seal ring stick out further as it looked too shallow and online data says it should be 13mm more than the pipe diameter.

Where a 3D difference is unavoidable, as in this case for the T junctions, then I difference just the pipes and render them. So the preview code has no 3D differences done by CGAL. This can get very slow when there are thousands of vitamins in one view.

I used the property accessor functions in the test routine instead of hard coded indexes. 

I replaced all the translate([0, 0, Z]) by translate_z(Z).

You might want to check it matches your parts in case I broke it.

PR #291, opened 2024-10-17
Add 10x15 micro rocker switches

mmuman:
Needed these tiny things.

Not sure about all measurements yet.

### DIFF
--- a/vitamins/rockers.scad
+++ b/vitamins/rockers.scad
@@ -19,6 +19,9 @@
 
 include <spades.scad>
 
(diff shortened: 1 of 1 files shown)


mmuman:
Fixed dimensions. Can't run tests for now as I don't fancy pulling ImageMagick 7 form debian/experimental yet, I'll have a look later.

nophead:
ImageMagick 7 has been around for more than 5 years I think. Not sure why it would need an experimental os version to get it.

mmuman:
Debian Sid still has some 6 something… probably because nobody stepped in to update it.
Thanks for merging in.

PR #298, opened 2025-03-02
Dimensions

AlexVerschoot:
I wanted to design a new piece for the library, based on an existing one, but i had trouble figuring out what each parameter meant. There's a tried and true solution for that : dimensions. 

However, i didn't really find a good way to add them, so I implemented them in 3d.
I also added a whole host of examples in the test folder :
image

I also re implemented  rotation matrix code in the math.scad file, because i didn't really find an elegant way to add them using the existing code.

I also didn't really add any parameters for group enabling/disabling them. I don't know if that's an idea that would be liked?

Best regards


nophead:
I think dimensions should be implemented to normal drawing standards. I.e. having extension lines and having the text breaking the line, not above it.

image

I used some on a project but was too lazy to add the arrow heads and probably not general enough to be included in the library.

rod_assembly

I don't understand why you needed new rotation code. The existing `rotate()` function produces a 4x4 rotation matrix and it can be applied to a vector using `transform()` or to geometry with `multmatrix()`.

AlexVerschoot:
> I think dimensions should be implemented to normal drawing standards. I.e. having extension lines and having the text breaking the line, not above it.

Normal drawing standards depend on your location. Me being a continental European, i defaulted to the ISO standards way of dimensioning, while yours is the ANSI way. 

See:
https://engineering.stackexchange.com/questions/45142/does-asme-follow-the-ansi-standard-the-iso-standard-or-neither-both-work
7VEiN


> I don't understand why you needed new rotation code. The existing rotate() function produces a 4x4 rotation matrix and it can be applied to a vector using transform() or to geometry with multmatrix().


I first wanted to gauge your interest in dimensions, and their inclusion into the library. The way i implemented them (in 3d), the function calls...
The math code was mostly code i had laying around, and knew how to use. I never used the NopSCADlib math functions and so went familiar code. 
If you think i'm wasting both our times with this, i'd rather know before i spend too much time on it. If you think this is something you'de be willing to include, i'll re implement it using the already included math code.

Also, it's currently very basic (enough for my needs to figure out some what some parametric dimensions do, and easy reference for 3d printing), but following items could still be included
- dimensions to the side (for short measurements)
- radius 
- tolerances
- toggle to switch between ANSI/ISO way
- inclusion of dimension into documentation of parameterized library items
- parameter for rotating them around the measurement axis
- parameter for text size
- plane options?
- font options?
- default to only being used in $preview mode?
- ...?

 
I wanted to gauge if you have a philosophy of "if someone needs it they will add it if they need it" or "I only add feature complete items".

nophead:
Sorry I wasn't aware there were two different standards.  I am British so I should use ISO as well. It is 44 years since I did technical drawing at school and I haven't done it since.

Additions are welcome and I think dimensions are useful, so you are certainly not wasting your time. And it doesn't need to be complete and include everything before it is merged, it just needs to be useful. The main issues are I think it should have the extension lines with a gap to the object and not have its own maths using 3x3 matrices instead of 4x4, which is the way most 3D libraries work.

This is the code I used:-
```
module dimv(p1, p2, z, max = inf, offset = 1) { //! Draw a vertical dimension showing the x,y distance between two points
    for(p = [p1, p2])
        translate(p + [0, 0, offset])
            cylinder(r = 0.1, h = z - p.z - offset);

    mid = (p1 + p2) / 2;
    d = norm(vec2(p1 - p2));
    text = str(d, "mm");
    tw = d - 2;
    if(max < tw)
        for(side = [-1, 1])
            translate([mid.x, mid.y, 0] + [side * d / 2, 0, z - max / len(text) / 1.75])
                rotate([0, -side * 90, 0])
                    cylinder(r = 0.1, h = (d - max) / 2 - 1);

    translate([mid.x, mid.y, z])
        rotate([90, 0, 0])
            resize([min(tw, max), 0], auto = true)
                linear_extrude(0.1)
                    text(text, halign = "center", valign = "top");
}
```

I guarded it with if(exploded()) so it only occurred on exploded assembly views, which themselves are only when $preview is set.

Not sure about 3D arrow heads. Perhaps they could be triangles orientated to be flat to the camera instead, using the the $vpr variable.

nophead:
Let me know when you think it is ready to merge.

AlexVerschoot:
Hi,

I think it's ready to be merged.
It still lacks following features :
- [ ] Dimensions to the side (for short measurements)
This one is kind of mitigated because of the standard automatic scaling, however, if i ever have some time left over, i will look into making it an option.
- [ ] Radius
While it's not the standard way, you can use the dimension and two points to make it clear. Should be implemented after dimensions to the side, because i think, at least for the iso way, they should use shared code.
- [ ] Tolerances
I mostly use this library for 3d printing, so not a priority for me.
- [ ] Toggle to switch between ANSI/ISO way
I'd actually love to use textmetrics() for this, but this is still in the development snapshot. So i'd wait until that has landed. I don't like to assume the size of the characters.

I also changed the code to do away with calculating the offsets myself, and instead use native openscad functions. I also changed the text to self-drawn polygon triangles at the end of the dim.

What did i add: 
- Dimension: a module to draw a dimension between any two points in a 3d space:
```
//if text is empty, will display the number value
//thickness will determine the thickness of the lines, and size of the arrows, if 0, will use 0.5% of the length of the dim
//text_size will determine the size of the text, if 0, will use percentage of the length of the dim
module dimension(startpoint, endpoint, text = "", thickness = 0, text_size = 0 , rot_around_dim=0)
```
afbeelding

- Dimension_x
A module to display a dimension in x :
```
//offset will determine how much space is between the measured point and the dimension
//for x, this offset will be in the y direction
//plane options : xy, xz
module dimension_x(startpoint, endpoint, offset = 1, text = "", thickness = 0, text_size = 0 , plane = "xy")
``` 
xy plane :
afbeelding
xz plane :
afbeelding

- Dimension_y
Similar to Dimension_x

- Dimension_z
Similar to Dimension_x

AlexVerschoot:
Changed it.

AlexVerschoot:
I changed the comment style, but while already using it, i found a critical bug where a negative offset didn't invert the direction of the dimension_x,y en z (lower instead of higher).

I will be fixing it first.

AlexVerschoot:
Hi, I changed that. Also made certain negative offsets work as I would expect.
One of the purposes of me doing this, was figuring out the t-nut parameters. And i think it works quite well for that:
```
include <NopSCADlib/core.scad>
include <NopSCADlib/lib.scad>
include <NopSCADlib/vitamins/nut.scad>
include <NopSCADlib/utils/dimension.scad>

/*[Dimensions]*/
//Show the paremeter value/name
params = false; 

screw_dia = 3;
dimension_y([0,-screw_dia/2,thickness], [0,screw_dia/2,thickness], offset = 1, text=(params?"screw_dia":""),plane = "yz");
nut_dia = 6;
dimension_y([0,-nut_dia/2,thickness-nyloc_thickness], [0,nut_dia/2,thickness-nyloc_thickness], offset = -1, text=(params?"nut_dia":""),plane = "yz");
thickness = 3.0;
dimension_z([0,ty2/2,thickness], [0,ty1/2,0], text=(params?"thickness":""),offset = 1, plane = "yz");

nyloc_thickness = 4.0;
dimension_z([0,ty1/2,thickness], [0,ty2/2,thickness-nyloc_thickness], text=(params?"nyloc_thickness":""),offset = 2, plane = "yz");

washer = false;

trap_depth = 0;
sx =10;
ty1=10;
ty2=6;
dimension_y([0,-ty1/2,0], [0,ty1/2,0], text=(params?"ty1":""),offset = thickness+3, plane = "yz");
dimension_y([0,-ty2/2,thickness], [0,ty2/2,thickness], text=(params?"ty2":""),offset = 2, plane = "yz");
//dimension_z([0,0], [], plane="yz");

hammer = false;


translate([0,0,0])
sliding_t_nut(["M3_sliding_t_nut", screw_dia,   nut_dia,   thickness, nyloc_thickness,  washer,         trap_depth,  sx,  ty1,  ty2, hammer]);
```
image

Is this kind of thing something i should add somewhere? If yes, where?

nophead:
You can put the image in the docs folder and then reference it in the documentation for dimensions. See printed/box.scad for an example.  You can put the source code inline with triple backticks.

AlexVerschoot:
That should do it, i added the picture to the nut tough, i figured it would be easier to find. I didn't know what you meant by the triple backticks. Would you mind merging it in so i can study what you did?

Best regards>

PR #305, opened 2025-04-14
positional parameters should not follow named parameters

schiele:
Some versions of OpenSCAD (like the latest official release 2021.01 get confused if positional parameters follow named parameters. So let's use named parameters after the first named one was used.


nophead:
Yes I agree, I will merge this when I have time.

schiele:
Fixed two more cases where this same rule was violated.

nophead:
Thanks

PR #306, opened 2025-04-22
[New Feature] Added link to STL files in readme.md

pcremades:
Hi,
I changed views.py so that it includes links to stl files in readme.md. This makes easy for people to download stl files from repo without havin to browse the stls folder.


nophead:
Sorry I missed this last week. Does it also work when the readme is hosted by Github?

pcremades:
Done. I made the same modification so that DXFs and SVGs are also linked in the documentation.

nophead:
I don't think is is correct for SVGs as it assumes the dxfs directory for routed parts but svgs are in the svgs directory,

pcremades:
Sorry, my bad. I made the necessary changes and tested with BOM.scad. It seems to be working ok now.

PR #318, opened 2026-03-13
Rename all identifers that start with a numeral

### LINKED ISSUE #307: [Deprecation] Variable names starting with digits
Hi all,
I'm trying version 2025.04.27 of OpenSCAD and I receive tons of warning related to variable names starting with digits in NopSCADlib. This will become a big issue.

JamesRandom:
To prevent warnings in latest versions of OpenSCAD all identifiers that start with a digit have an undersoce prefix added.

Fixes: #307


jeroenrnl:
I personally find using underscores at the start of a variable a rather nasty hack. Also, in some languages that don't have private variables or functions, it's common to use and underscore as the first character to indicate this is an internal/private method, function or variable. 
I think we could do better, e.g. `2p54header` -> `header_2p54`, `7_segment_digits` -> `seven_segment_digits` or `display_7_segments` and something similar could probably be found for all/most of them. I'd be happy to implement this but would like @nophead 's opinon on this. (also, I think his approach of "simply not upgrading OpenSCAD anymore" is not feasible in the long run)

nophead:
NopSCADlib only works properly with the last release and a few snapshots later due to a change to the semantics of $variables. I don't want to change all my projects and I am terminally ill, so I have better things to do.

